PR #36, opened 2024-09-23
CP-22073: Remove Service Discovery for Node Exporter

bdrennz:
This PR removes references to Node Exporter as this service is no longer used by the [Prometheus Agent](https://github.com/Cloudzero/cloudzero-charts).

Accompanying commit to remove node-exporter from the chart: https://github.com/Cloudzero/cloudzero-charts/pull/78/commits/1562e94692d3591c335a80521342bdca5ed51703


pkg/cmd/config/internal/scrape_config.tmpl
```diff
@@ -7,63 +7,6 @@ global:
   - PrometheusText0.0.4
   evaluation_interval: 1m
 scrape_configs:
-- job_name: cloudzero-nodes-cadvisor
```
bdrennz, line 10: Would appreciate a sanity check here. My understanding is the [KSM section](https://github.com/Cloudzero/cloudzero-agent-validator/blob/develop/pkg/cmd/config/internal/scrape_config.tmpl#L67) is the only scrape config needed.

dmepham: I don't think we want to get rid of this one. this gets the `container_` metrics. I think we still get the `node_` metric(s) from the other job. I think the change on the chart side would be to not deploy node exporter I guess, and also remove `node_dmi_info` from the values file.

dmepham (approved):

josephbarnett (approved):

PR #39, opened 2024-09-30
CP-21862: Generate Command should Allow users to specify CloudZero Secret File Path

bdrennz:
This PR modifies the `generate` command by allowing users to include the path to their secret. 

The value originates from: https://github.com/Cloudzero/cloudzero-charts/blob/develop/charts/cloudzero-agent/templates/cm.yaml#L162

### DIFF
--- a/pkg/cmd/config/command.go
+++ b/pkg/cmd/config/command.go
@@ -29,6 +29,7 @@ type ScrapeConfigData struct {
 	CloudAccountID string
 	Region         string
 	Host           string
+	SecretPath     string
 }
 
 func NewCommand(ctx context.Context) *cli.Command {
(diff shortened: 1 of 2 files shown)


pkg/cmd/config/command.go
```diff
@@ -48,12 +49,14 @@ func NewCommand(ctx context.Context) *cli.Command {
 					&cli.StringFlag{Name: "configmap", Usage: "name of the ConfigMap", Required: true},
 					&cli.StringFlag{Name: "pod", Usage: "name of the cloudzero-agent pod", Required: true},
 					&cli.StringFlag{Name: "host", Usage: "host for the prometheus remote write endpoint", Required: true},
+					&cli.StringFlag{Name: "secret-path", Usage: "path to the secret file", Value: "/etc/config/prometheus/secrets/", Required: false},
```
dmepham, line 52: I think being compatible with the existing helm template would require a field for the secret file name as well, which the chart allows users to choose using `containerSecretFileName` right now. this is because some users might be using secrets providers that have specific names for files/paths that we can't control

bdrennz: I think I'm following. If we extract the value from https://github.com/Cloudzero/cloudzero-charts/blob/develop/charts/cloudzero-agent/templates/_helpers.tpl#L21, then we should be okay since that contains the file name as well, right?

bdrennz: Here's the accompanying commit on the chart side: https://github.com/Cloudzero/cloudzero-charts/pull/78/commits/48c4f0e2e049a42b14f1abf9fedc21b8357ede99

dmepham: ah perfect, makes sense given the chart commit - thanks!

bdrennz: Thanks for keeping me honest!

PR #40, opened 2024-10-02
CP-22291: remove latest tag in build

dmepham:
I changed `latest=auto` to `latest=false` based on the docs [here](https://github.com/docker/metadata-action/tree/8e5442c4ef9f78752691e2d8f8d19755c6f78e81/?tab=readme-ov-file#flavor-input), with the thinking that we don't want this to be handled by this build step

testing on my fork in progress


adamrice32 (approved):
Joe approves. ;)

Suggestion - merge and test immediately. Be prepared to update latest tag by hand.

PR #50, opened 2024-10-15
CP-22509: Search All Namespaces for KSM and Improve Search Logic

bdrennz:
This PR modifies the `generate` command to search for KSM in _all_ namespaces. Additionally, if KSM is not found by searching for the substring in the Service name iteself, the command checks the Service's Labels for `"app.kubernetes.io/name": "kube-state-metrics`, which should be more robust.

### DIFF
--- a/pkg/k8s/services.go
+++ b/pkg/k8s/services.go
@@ -7,8 +7,8 @@ import (
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 )
@@ -53,9 +53,10 @@ func BuildKubeClient(kubeconfigPath string) (kubernetes.Interface, error) {
 	return clientset, nil
 }
 
-// GetKubeStateMetricsURL fetches the URL for the Kube State Metrics service
-func GetKubeStateMetricsURL(ctx context.Context, clientset kubernetes.Interface, namespace string) (string, error) {
-	services, err := clientset.CoreV1().Services(namespace).List(ctx, metav1.ListOptions{})
+// GetKubeStateMetricsURL fetches the URL for the Kube State Metrics service across all namespaces
+func GetKubeStateMetricsURL(ctx context.Context, clientset kubernetes.Interface) (string, error) {
+	// First, try to find the service by name
+	services, err := clientset.CoreV1().Services("").List(ctx, metav1.ListOptions{})
 	if err != nil {
 		return "", errors.Wrap(err, "listing services")
 	}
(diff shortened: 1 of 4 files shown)


pkg/k8s/services.go
```diff
@@ -65,7 +66,17 @@ func GetKubeStateMetricsURL(ctx context.Context, clientset kubernetes.Interface,
 	for _, service := range services.Items {
 		if strings.Contains(service.Name, "kube-state-metrics") {
 			kubeStateMetricsURL = fmt.Sprintf("%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, service.Spec.Ports[0].Port)
-			break
+			return kubeStateMetricsURL, nil
+		}
+	}
+
+	// If not found by name, check by labels
+	for _, service := range services.Items {
```
bdrennz, line 74: This is the additional check for the KSM Label

pkg/k8s/services.go
```diff
@@ -53,9 +53,10 @@ func BuildKubeClient(kubeconfigPath string) (kubernetes.Interface, error) {
 	return clientset, nil
 }
 
-// GetKubeStateMetricsURL fetches the URL for the Kube State Metrics service
-func GetKubeStateMetricsURL(ctx context.Context, clientset kubernetes.Interface, namespace string) (string, error) {
-	services, err := clientset.CoreV1().Services(namespace).List(ctx, metav1.ListOptions{})
+// GetKubeStateMetricsURL fetches the URL for the Kube State Metrics service across all namespaces
+func GetKubeStateMetricsURL(ctx context.Context, clientset kubernetes.Interface) (string, error) {
+	// First, try to find the service by name
+	services, err := clientset.CoreV1().Services("").List(ctx, metav1.ListOptions{})
```
bdrennz, line 59: By setting the `namespace` arg to `""`, the function call will return all services across all namespaces.

dmepham (approved):
lgtm - though one question I have is, what should the `kube_state_metrics_service_endpoint` config value represent now? as far as I can tell, that config value is still required and the `Validate` method still validates that it is a valid endpoint. if we're searching for the KSM endpoint in all namespaces, is this not needed?

one thing we could do is still allow that config value, and let it be an override for the KSM endpoint we check for. this would handle the case where someone wanted to use an existing KSM, but it didn't conform to the naming we expect.

pkg/k8s/services.go
```diff
@@ -65,7 +66,17 @@ func GetKubeStateMetricsURL(ctx context.Context, clientset kubernetes.Interface,
...
 			kubeStateMetricsURL = fmt.Sprintf("%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, service.Spec.Ports[0].Port)
-			break
+			return kubeStateMetricsURL, nil
+		}
+	}
+
+	// If not found by name, check by labels
+	for _, service := range services.Items {
+		// Check for Helm-specific labels
+		if service.Labels["app.kubernetes.io/name"] == "kube-state-metrics" &&
```
josephbarnett, line 76: small nit - you might move these up to the top of the file as a const.

josephbarnett (approved):
I agree w/ Dans comment - kms parameter for validate is less useful not, and improves the chart configuration story. I'd recommend a followup for that tho.

bdrennz:
@dmepham The value still represents what it did previously did. I thought about allowing users to still override this value, but that sort of defeats the purpose of the feature in my oppinion.

Ultimately there's two scenarios we'll see:

1. User enabled KSM and allows the Chart to create the service. At which point, the [validator config map](https://github.com/Cloudzero/cloudzero-charts/pull/78/files#diff-f4b4daf22c6474bc2144d2622cadd1636f32c99045cfd1284236ab5669016f80R33) will route that to the expected endpoint .

2. The user has KSM configured and the validator needs to discover it. The only way for the validator _not_ to discover the endpoint would be if:
   a. The KSM service doesn't have `kube-state-metrics` in the service name.
   b. The KSM service doesn't have the  `"app.kubernetes.io/name": "kube-state-metrics` label.
   
 If for scenario `2.` , the validator fails to find the service because one of the two conditional checks fail, we can have them manually add the label to the service:
 ``` bash
 kubectl label service <service-name> app.kubernetes.io/name=kube-state-metrics
 ```

dmepham:
for case `#1` - maybe I'm missing it, because I'm not super familiar with this repo, but it looked like `kube_state_metrics_service_endpoint` is only used for validation? so, in the case where the user uses our KSM, it will still do the search for the KSM endpoint. which is fine, it will still find it. but it's just a case of having a chart/user input for validation but then dynamically finding the endpoint for the scrape config. but again, I don't know this code very well, so def feel free to disregard if I'm off base here.

and, in case `#2`, is [this address](https://github.com/Cloudzero/cloudzero-charts/pull/78/files#diff-f4b4daf22c6474bc2144d2622cadd1636f32c99045cfd1284236ab5669016f80R33)  correct if we're not using our own KSM? maybe that's a question for another PR, but if that `kube_state_metrics_service_endpoint` setting is used for validation, and the KSM address is different than that setting, won't validation fail?

aside from that, the only other reason that allowing a user to override the KSM with a manual input is if a cluster has a huge amount of services and it took some time to search. but, that's probably unlikely and a premature optimization at best. 

having said all that, looks like a good change, and I'm all good with merging this. just wanted to clarify those things for my own understanding

PR #64, opened 2024-11-27
CP-23050: Validate CloudZero Metrics are available from Kube State Metrics

bdrennz:
This PR modifies the `kube_state_metrics_reachable` `post-start` job to also validate that all CloudZero KMS metrics are available. A sample log output from the Job:
``` json
{
  "level": "info",
  "log_sequence": 6,
  "msg": "Using endpoint URL: http://192.168.5.249:8080/metrics",     <---- Internal IP discovered using the Service Endpoint 
  "op": "ksm",
  "time": "2024-11-25T08:23:44Z"
}
{
  "level": "info",
  "log_sequence": 7,
  "msg": "All required metrics found: [kube_pod_info kube_node_info]",                <---- Metric Check (subset for now)
  "op": "ksm",
 }
 
```

``` bash
Part of the Cluster Status Report related to KMS produced by the Job: {\"name\":\"kube_state_metrics_reachable\",\"passing\":true}]}",
```

The log statements show that the job was able to discover the internal IP of the KSM service, find all required metrics, and mark the `kube_state_metrics_reachable` job as complete.

It's worth noting that the base branch is `feature/cp-23050`, not `develop`. The remaining work will be handled in a subsequent ticket. I hard coded some values temporarily to prove out the functionality until we can get around to the Chart changes in https://cloudzero.atlassian.net/browse/CP-23740.

### DIFF
--- a/pkg/diagnostic/kms/check.go
+++ b/pkg/diagnostic/kms/check.go
@@ -73,32 +73,40 @@ func (c *checker) Check(ctx context.Context, client *http.Client, accessor statu
 	// Retry logic to handle transient issues
 	for attempt := 1; retriesRemaining > 0; attempt++ {
 		resp, err := client.Get(endpointURL)
-		if err == nil && resp.StatusCode == http.StatusOK {
-			defer resp.Body.Close()
-			body, err := io.ReadAll(resp.Body)
-			if err != nil {
-				c.logger.Errorf("Failed to read metrics on attempt %d: %v", attempt, err)
-				accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Failed to read metrics: %s", err.Error())})
-				return nil
-			}
+		if err != nil {
+			c.logger.Errorf("Failed to fetch metrics on attempt %d: %v", attempt, err)
+			retriesRemaining--
+			time.Sleep(RetryInterval)
+			continue
+		}
 
-			metrics := string(body)
-			requiredMetrics := []string{"kube_pod_info", "kube_node_info"} // Add the required metrics here
-			for _, metric := range requiredMetrics {
-				if !strings.Contains(metrics, metric) {
-					c.logger.Errorf("Required metric %s not found on attempt %d", metric, attempt)
-					accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Required metric %s not found", metric)})
-					return nil
-				}
-			}
+		if resp.StatusCode != http.StatusOK {
+			c.logger.Errorf("Unexpected status code on attempt %d: %d", attempt, resp.StatusCode)
+			retriesRemaining--
+			time.Sleep(RetryInterval)
+			continue
+		}
 
-			accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: true})
+		defer resp.Body.Close()
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			c.logger.Errorf("Failed to read metrics on attempt %d: %v", attempt, err)
+			accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Failed to read metrics: %s", err.Error())})
 			return nil
 		}
 
-		c.logger.Errorf("Failed to fetch metrics on attempt %d: %v", attempt, err)
-		retriesRemaining--
-		time.Sleep(RetryInterval)
+		metrics := string(body)
+		requiredMetrics := []string{"kube_pod_info", "kube_node_info"} // Add the required metrics here
+		for _, metric := range requiredMetrics {
+			if !strings.Contains(metrics, metric) {
+				c.logger.Errorf("Required metric %s not found on attempt %d", metric, attempt)
+				accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Required metric %s not found", metric)})
+				return nil
+			}
+		}
+
+		accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: true})
+		return nil
 	}
 
 	accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Failed to fetch metrics after %d retries", MaxRetry)})
--- a/pkg/diagnostic/kms/check.go
+++ b/pkg/diagnostic/kms/check.go
@@ -62,7 +62,7 @@ var NewProvider = func(ctx context.Context, cfg *config.Settings, clientset ...k
 	}
 }
 
-func (c *checker) Check(ctx context.Context, client *http.Client, accessor status.Accessor) error {
+func (c *checker) Check(_ context.Context, client *http.Client, accessor status.Accessor) error {
 	var (
 		retriesRemaining = MaxRetry
 		endpointURL      = fmt.Sprintf("%s/metrics", c.cfg.Prometheus.KubeStateMetricsServiceEndpoint)
--- a/pkg/diagnostic/kms/check.go
+++ b/pkg/diagnostic/kms/check.go
@@ -3,15 +3,20 @@ package kms
 import (
 	"context"
 	"fmt"
-	net "net/http"
+	"io/ioutil"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/cloudzero/cloudzero-agent-validator/pkg/config"
 	"github.com/cloudzero/cloudzero-agent-validator/pkg/diagnostic"
-	"github.com/cloudzero/cloudzero-agent-validator/pkg/http"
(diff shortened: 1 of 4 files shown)


pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
 	}
 }
 
-func (c *checker) Check(ctx context.Context, client *net.Client, accessor status.Accessor) error {
+func (c *checker) Check(ctx context.Context, client *http.Client, accessor status.Accessor) error {
 	var (
-		err              error
 		retriesRemaining = MaxRetry
-		url              = fmt.Sprintf("%s/", c.cfg.Prometheus.KubeStateMetricsServiceEndpoint)
+		namespace        = "prom-agent"
```
bdrennz, line 69: note: should be passed by the Agent Chart. Will be handled in: https://cloudzero.atlassian.net/browse/CP-23740

bdrennz, line 69: note: should be passed by the Agent Chart. Will be handled in: https://cloudzero.atlassian.net/browse/CP-23740

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
+		serviceName      = "cz-prom-agent-kube-state-metrics"
```
bdrennz, line 70: Note: `serviceName` should be passed by the Agent Chart. Will be handled in: https://cloudzero.atlassian.net/browse/CP-23740

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
+		// Log the endpoints for debugging
+		c.logger.Infof("Endpoints: %v", endpoints)
+
+		// Check if there are any ready addresses and find the first available endpoint
+		for _, subset := range endpoints.Subsets {
+			for _, address := range subset.Addresses {
+				c.logger.Infof("Address: %v", address)
+				for _, port := range subset.Ports {
+					c.logger.Infof("Port: %v", port)
+					if port.Port == 8080 {
```
bdrennz, line 92: Note: `port` should be passed by the Agent Chart. Will be handled in: https://cloudzero.atlassian.net/browse/CP-23740

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
+			defer resp.Body.Close()
+			body, err := io.ReadAll(resp.Body)
+			if err != nil {
+				c.logger.Errorf("Failed to read metrics: %v", err)
+				accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Failed to read metrics: %s", err.Error())})
+				return nil
+			}
+
+			metrics := string(body)
+			requiredMetrics := []string{"kube_pod_info", "kube_node_info"} // Add the required metrics here
```
bdrennz, line 137: Note: `requiredMetrics` should be passed by the Agent Chart. Will be handled in: https://cloudzero.atlassian.net/browse/CP-23740

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
-	logger *logrus.Entry
+	cfg       *config.Settings
+	logger    *logrus.Entry
+	clientset kubernetes.Interface
 }
 
-func NewProvider(ctx context.Context, cfg *config.Settings) diagnostic.Provider {
+var NewProvider = func(ctx context.Context, cfg *config.Settings, clientset ...kubernetes.Interface) diagnostic.Provider {
+	var cs kubernetes.Interface
+	if len(clientset) > 0 {
```
dmepham, line 38: just curious, why do we need this?

pkg/diagnostic/kms/check.go
dmepham: is there a PR to update the chart to pass these in?

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
+			accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Failed to get service endpoints: %s", err.Error())})
+			return nil
 		}
-		if retriesRemaining == 0 {
+
+		// Log the endpoints for debugging
+		c.logger.Infof("Endpoints: %v", endpoints)
+
+		// Check if there are any ready addresses and find the first available endpoint
+		for _, subset := range endpoints.Subsets {
```
dmepham, line 87: I don't think I follow - why do we need to find the address and port? if we're managing KSM, we'll know what the dns and port will be, right? why not just check the given Service address?

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
 
-	accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: true})
-	return nil
+	c.logger.Infof("Using endpoint URL: %s", endpointURL)
+
+	// Retry logic to handle transient issues
+	retriesRemaining = MaxRetry
+	for retriesRemaining > 0 {
+		resp, err := client.Get(endpointURL)
+		if err == nil && resp.StatusCode == http.StatusOK {
```
dmepham, line 127: nit: it might be a bit nicer and avoid some indentation if you check `err != nil ...` and `continue`

pkg/diagnostic/kms/check.go
```diff
@@ -23,46 +28,130 @@ var (
...
+					accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: false, Error: fmt.Sprintf("Required metric %s not found", metric)})
+					return nil
+				}
+			}
+
+			accessor.AddCheck(&status.StatusCheck{Name: DiagnosticKMS, Passing: true})
+			return nil
+		}
 
+		c.logger.Errorf("Failed to fetch metrics: %v", err)
```
dmepham, line 150: nit: might be nice to note in this message that the request is being retried

dmepham (changes requested):
lgtm! I have some clarifying questions, mostly about ordering of changes, as it looks like this requires some chart changes

pkg/diagnostic/kms/check.go
bdrennz: Not yet - I just created the ticket mentioned above as a placeholder.

pkg/diagnostic/kms/check.go
roberthocking, line 1: small nit - should this be `ksm` instead of `kms`

roberthocking (changes requested):
Looks Good, just have the question about `ksm` vs `kms`

pkg/diagnostic/kms/check.go
bdrennz: Yea that's just things were before

pkg/diagnostic/kms/check.go
bdrennz: `clientset` is an optional arg, mainly used for unit tests. This is just a way of checking that one is being passed and if so, prefer that over finding it dynamically based on the `InClusterConfig`. For a bit of background, mocking the `InClusterConfig` proved to be difficult, so this was an easy alternative.

pkg/diagnostic/kms/check.go
dmepham: > The Service Endpoint provided by the Chart doesn't actually work from within the container

meaning, you can't hit `http://<release-name>.<release-namesapce>.svc.cluster.local` from within the container? that's surprising to me. that's how I generally address other pods within a cluster. for example, [here](https://github.com/Cloudzero/cloudzero-charts/blob/feature/1.0.2-beta-release/charts/cloudzero-insights-controller/templates/init-job.yaml#L28)

> There's not an easy way to have that value statically defined and passed from the chart.

isn't the KSM service name always `cloudzero-state-metrics`? because of [this](https://github.com/Cloudzero/cloudzero-charts/blob/feature/1.0.2-beta-release/charts/cloudzero-agent/values.yaml#L65-L66) override. and then we already know the namespace that the chart is deployed in. so I would think you'd be able to put the following in the validator config:

```yaml
kube_state_metrics_endpoint: https://cloudzero-state-metrics.{{ .Release.Namespace }}.svc.cluster.local:8080
```
or, you could probably get the svc name and port dynamically as well:
```yaml
kube_state_metrics_endpoint: https://{{ .Values.kubeStateMetrics.nameOverride }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.kubeStateMetrics.service.port }}
```

I might be missing the point

bdrennz: Yea this seems right to me. What through me off with the service endpoint was the off formatting that the Agent Chart currently passes to the Validator. I tried with the formatting you mentioned above and it worked, I'll make the change. Thanks!

dmepham: yeah for sure! yeah it's a weird thing to get the name from the ksm subchart, but I'm glad it works!

pkg/diagnostic/kms/check.go
bdrennz: Cool if I tackle this in the next PR?

dmepham (approved):
lgtm!

PR #79, opened 2025-01-06
CP-24501: add cloudzero-agent-inpector

evan-cz:
This adds the inspector code as a separate binary (`cloudzero-agent-inpector`). It is installed into the same Docker image as the `cloudzero-agent-validator` binary.

There are a few tweaks around the build system, too... I changed the Dockerfile to run `make build` so we don't duplicate the build command, and added a `package-build` target to build the image but only send it to the local Docker install instead of pushing it out to the repo, which makes testing easier.

There are lots of unit tests, but I've also tested by building the image (using the new `package-build` target), then bringing it up in a local cluster with a tweaked version of the Helm chart from `cloudzero-charts` (see the [CP-24043 branch in `cloudzero-charts`](https://github.com/Cloudzero/cloudzero-charts/tree/CP-24043)). I have a patch ready to go for `cloudzero-charts` to integrate this work, but it won't really work until there has been a release of this repo.

I should also point out one deficiency, which is that the inspector logs using zerolog, not the logrus-based code in `pkg/logging`. That's mainly just because the inspector was developed independently, but I'm disinclined to switch it over right now because it seems like we might want to switch the c-a-v code instead (the insights controller is also using zerolog). Once we decide what to go with I'm happy to put together a follow-up patch to implement that, but IMHO that shouldn't block this PR.


docker/Dockerfile
```diff
@@ -60,8 +52,9 @@ VOLUME [ "/app/config" ]
...
 
 # Copy the Go binary from the builder stage
+COPY --from=builder /go/bin/cloudzero-agent-inspector /app/cloudzero-agent-inspector
 COPY --from=builder /go/bin/cloudzero-agent-validator /app/cloudzero-agent-validator
 
 # Allow the default ENTRYPOINT from busybox to be the default,
 # however run the app as the default command
-CMD ["/app/cloudzero-agent-validator", "-h"]
\ No newline at end of file
+CMD ["/app/cloudzero-agent-validator", "-h"]
```
dmepham, line 60: so by default, do we not start the inspector?

dmepham (approved):

docker/Dockerfile
evan-cz: Correct. We don't really start the validator either by default, just print the help screen; the entry point is overridden in the Helm chart to actually run the validator.

FWIW, this line isn't really changed, I just added a newline before the EOF.

The only thing this really does to the image is adds the `cloudzero-agent-inspector` binary alongside of `cloudzero-agent-validator`. Then, the validator container in the Helm chart runs `/app/cloudzero-agent-validator` and the (new, disabled by default) inspector container runs `/app/cloudzero-agent-inspector`.

dmepham: great, thanks1

PR #124, opened 2025-04-02
Chart: Add dnsPolicy and dnsConfig configurables to deployment/job specs

jalev:
Users might have some interesting DNS requirements (e.g. Wise). This allows them to override the `dnsConfig` and `dnsPolicy` without directly modifying the chart. 

Defaults are defined [as per Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy).

### DIFF
--- a/helm/templates/aggregator-deploy.yaml
+++ b/helm/templates/aggregator-deploy.yaml
@@ -101,7 +101,10 @@ spec:
         runAsNonRoot: true
         runAsGroup: 65534
         fsGroup: 65534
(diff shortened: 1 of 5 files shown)


evan-cz (approved):
Looks good to me. Thanks for moving the PR to this repo, and for your contribution!

I'll take it from here. Looks like something might be preventing the status checks from running for external contributions, if I can't get that sorted out quickly I'll just pull this into a branch and run it from there.

Thanks again!

PR #125, opened 2025-04-02
Add dnsPolicy and dnsConfig configurables to deployment/job specs

evan-cz:
This is just a copy of #124 to get the status checks to run. Figuring out why they aren't running for external contributions is a tomorrow problem...


evan-cz:
Hm, looks like the status checks ran when I pushed the change to a branch. I'm a little surprised GitHub was able to associate them with the original PR, I guess because it associates them with a commit ID (which does match)... I'll try to play with this at some point from my personal account.

PR #126, opened 2025-04-03
CP-27611: unify code for referencing images

evan-cz:
This adds an agent.images section, which contains default values for all images used in the Helm chart, as well as a helper function which accepts an image struct (which we use in many places) and falls back on the defaults from agent.images if a property isn't set.

Once we have made a release from the agent-validator repo, we should be able to delete *most* of the values from the image configuration sections and have them all use the default values automatically, though overrides set by customers will still be honored.


helm/values.yaml
```diff
@@ -405,6 +418,7 @@ kubeStateMetrics:
     registry: registry.k8s.io
     repository: kube-state-metrics/kube-state-metrics
     tag: "v2.10.1"
+    revision:
```
dmepham, line 421: is this used? I don't think `revision` is used in the KSM chart, and I don't think it's used in the chart

helm/values.yaml
```diff
@@ -120,6 +120,17 @@ prometheusMetrics:
 # -- Any items added to this array will be added to the metrics that are sent to CloudZero, in addition to the minimal labels that CloudZero requires.
 additionalMetricLabels: []
 
+# Agent largely contains top-level settings which are often shared by multiple
+# components within this chart, or used as defaults in case values are not
+# explicitly set per-component.
+agent:
+  # The Docker image used for most CloudZero components.
```
dmepham, line 127: nit of all nits: `Docker image` -> `container image`

helm/.gitignore
```diff
@@ -0,0 +1,2 @@
+/charts
```
dmepham, line 1: I don't quite remember, does this mean users will need to run a `helm dependency update` when installing if we don't bundle this ourselves? do argocd and others rely on the dependent charts already being bundled?

helm/templates/_helpers.tpl
```diff
@@ -405,23 +405,6 @@ imagePullSecrets:
 {{- end }}
 {{- end }}
 
-
-{{/*
-Get the full container image reference for the init scrape job pod
-*/}}
-{{- define "cloudzero-agent.initBackfillJob.imageReference" -}}
-{{ $backFillValues := (include "cloudzero-agent.backFill" .) | fromYaml }}
```
dmepham, line 413: I think this is fine to remove, but just noting that this particular line was trying to maintain some backwards compatibility from back when this was called `scrape` and not `backfill`. but, that was only in 1.0 beta versions, so I think it's good to not have that anymore?

helm/values.yaml
```diff
@@ -368,10 +379,11 @@ serverConfig:
...
   # imagePullSecrets: []
-  # image:
-  #   repository: ghcr.io/cloudzero/cloudzero-insights-controller/cloudzero-insights-controller
-  #   tag: 0.1.1
-  #   pullPolicy: Always
+  image:
+    repository:
+    tag:
+    revision:
+    pullPolicy:
```
dmepham, line 386: is it more idiomatic to comment these out, or leave them empty? I don't have a preference either way

dmepham (approved):
I have a couple nit questions, but looks great!

helm/values.yaml
evan-cz: Yes, sorry, that should be "digest" not revision... I'll check them all.

helm/.gitignore
evan-cz: We didn't bundle them before. There is actually a [.gitignore line in the charts repo](https://github.com/Cloudzero/cloudzero-charts/blob/88fe9ed67bddc98ce346c07aca393101c302fa57/.gitignore#L146) for this, it just doesn't effect this repo since it's in the top-level .gitignore in cloudzero-charts.

dmepham: oh gotcha, thanks!

helm/templates/_helpers.tpl
evan-cz: Yeah, breaking BC with a beta isn't a problem at all.

helm/values.yaml
dmepham: sounds good to me!

helm/values.yaml
```diff
@@ -120,6 +120,17 @@ prometheusMetrics:
 # -- Any items added to this array will be added to the metrics that are sent to CloudZero, in addition to the minimal labels that CloudZero requires.
 additionalMetricLabels: []
 
+# Agent largely contains top-level settings which are often shared by multiple
+# components within this chart, or used as defaults in case values are not
+# explicitly set per-component.
+agent:
+  # The Docker image used for most CloudZero components.
+  image:
+    repository: ghcr.io/cloudzero/cloudzero-agent-validator/cloudzero-agent-validator
```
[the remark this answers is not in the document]
evan-cz: I don't think we need to bother; once we make a beta-3 release from this repo (after renaming it) we can update this line and remove the values from a bunch of other places so they fall back on this. Until then, though, we're still overriding this in most places. to use insights-controller.

PR #128, opened 2025-04-03
Initial k8s test matrix PoC

spkane:
This is a GitHub Action that can run something against multiple releases of Kubernetes.

At the moment it is using `k3s` and installing the agent helm chart and is working with `act` locally.


.github/workflows/test-matrix-k8s-k3s.yaml
```diff
@@ -0,0 +1,57 @@
...
+          echo "Applying Helm Chart"
+          helm repo add prom https://prometheus-community.github.io/helm-charts
+          helm dependency build ./helm
+          export CLUSTERNAME="$(hostname)-$(date '+%Y%m%d%H%M%S')"
+          echo $CLUSTERNAME
+          helm install -n cza --create-namespace cza ./helm \
+            --set cloudAccountId="gha-0001" \
+            --set clusterName="${CLUSTERNAME}" \
+            --set region="gha" \
+            --set apiKey=${{ secrets.CLOUDZERO_DEV_API_KEY }} \
```
evan-cz, line 52: We'll need `--set host=dev-api.cloudzero.com` too, since it's a dev key.

I wouldn't mind integrating this command into the `Makefile`, so we could just do `make helm-install` or something (probably using `helm upgrade --install` instead of `helm install`). We already have the relevant variables set ([Makefile:198-201](https://github.com/Cloudzero/cloudzero-agent-validator/blob/160183621310aee6b58da7a47630879368f1143e/Makefile#L198-L201)), and I hate the idea of duplicating logic.

Completely untested, but something along the lines of

```make
PROMETHEUS_COMMUNITY_REPO ?= 'https://prometheus-community.github.io/helm-charts'
HELM_TARGET_NAMESPACE     ?= cz-agent
HELM_TARGET               ?= cz-agent
HELM                      ?= helm

.PHONY: helm-install
helm-install: api-tests-check-env
helm-install: ## Install the Helm chart
  $(HELM) repo list | grep "$(PROMETHEUS_COMMUNITY_REPO)" || $(HELM) repo add prometheus-community "$(PROMETHEUS_COMMUNITY_REPO)"
  $(HELM) dependency build ./helm
  @$(HELM) upgrade --install \
    -n "$(HELM_TARGET_NAMESPACE)" \
    --create-namespace "$(HELM_TARGET)" \
    ./helm \
    --set cloudAccountId="$(CLOUD_ACCOUNT_ID)" \
    --set clusterName="$(CLUSTER_NAME)" \
    --set region="$(CSP_REGION)" \
    --set apiKey="$(CLOUDZERO_DEV_API_KEY)"

.PHONY: helm-uninstall
helm-uninstall: ## Uninstall the Helm chart
  $(HELM) uninstall -n "$(HELM_TARGET_NAMESPACE)" "$(HELM_TARGET)"
```

.gitignore
```diff
@@ -1,4 +1,6 @@
 # local development tailoring
+*ignore*
+.secrets
```
evan-cz, line 3: What is `.secrets`? If it's just somewhere to store your API key, `local-config.mk` would be appropriate; it is included in the Makefile, and as such is a convenient place to set `CLOUDZERO_DEV_API_KEY`.

spkane: `.secrets` is the default file that the `act` tool looks for to populate secrets that would normally be provided by GitHub, when testing locally. That being said, I'll take a look at `local-config.mk`. As long as it is uses the dotEnv (`.env`) format it should be compatible, and could be used with `act` by providing a simple command line argument.

- https://github.com/nektos/act/discussions/1439
- https://nektosact.com/usage/index.html#secrets

.github/workflows/test-matrix-k8s-k3s.yaml
spkane: Ok. I'll take a look at adding something along these lines.

.gitignore
evan-cz: Ah, okay.

`local-config.mk` is a Makefile, so I guess technically it would be a superset of `.env`. It's intentional that there isn't a `local-config.mk` in the repository (it's for local configuration overrides), so I suspect you probably *could* do it.

I could also see adding a make target:

```make
.secrets:
  @echo "CLOUDZERO_DEV_API_KEY=\"$(CLOUDZERO_DEV_API_KEY)\"" > "$@"
```

If you do decide to use `.secrets`, the line in .gitignore should probably be `/.secrets` not `.secrets` so it only ignores the file in the top-level, not any subdirectory.

FWIW I don't love the `*ignore*` line either; I can *very* easily see that matching some files we really don't want it to. For example, maybe an `ignore_list.go` or something... I'm not sure what exactly your thinking with that line is, but I suspect it could be scoped a lot tighter too.

spkane: > Ah, okay.
> 
> `local-config.mk` is a Makefile, so I guess technically it would be a superset of `.env`. It's intentional that there isn't a `local-config.mk` in the repository (it's for local configuration overrides), so I suspect you probably _could_ do it.
> 
> I could also see adding a make target:
> 
> ```makefile
> .secrets:
>   @echo "CLOUDZERO_DEV_API_KEY=\"$(CLOUDZERO_DEV_API_KEY)\"" > "$@"
> ```

The Makefile target is probably a good approach to solving this while keeping the workflow reasonably straight forward and repeatable.

> If you do decide to use `.secrets`, the line in .gitignore should probably be `/.secrets` not `.secrets` so it only ignores the file in the top-level, not any subdirectory.

Actually, it belongs in the `.github/workflows/` directory by default, but agreed that once it is ready for a merge being specific is better.

> FWIW I don't love the `*ignore*` line either; I can _very_ easily see that matching some files we really don't want it to. For example, maybe an `ignore_list.go` or something... I'm not sure what exactly your thinking with that line is, but I suspect it could be scoped a lot tighter too.

Agreed. It was just a quick solution for the in-progress draft PR. I personally like having some pattern in the  `.gitignore` which makes it easy to drop a file into a repo dir for temporary notes, etc. that I know won't get committed, but it can be anything and does not really need to be there at all either. However, if used, it could be very specific as well like `notes-local-private.md` or whatever, it doesn't really matter as long as it is known.

evan-cz: > Agreed. It was just a quick solution for the in-progress draft PR. I personally like having some pattern in the `.gitignore` which makes it easy to drop a file into a repo dir for temporary notes, etc. that I know won't get committed, but it can be anything and does not really need to be there at all either. However, if used, it could be very specific as well like `notes-local-private.md` or whatever, it doesn't really matter as long as it is known.

How about a whole directory? `/scratch/`, perhaps?

I actually have a `scratch` directory that I use like that already, but I didn't bother adding a .gitignore for it. That said, I'm pretty careful about what I added (I always use `git add -p` and go through every change), which I don't think is the norm.

dmepham (approved):
this looks great!

josephbarnett (approved):
Nice! 💅🏻

PR #136, opened 2025-04-07
Cp 27052 improve http

jake-cloudzero:


evan-cz (commented):
Generally looks good, but we already have this code in the repo under pkg/inspector. Moving to app/ is cool (we want to do that anyways), or just using the copy in pkg/inspector, but we definitely shouldn't duplicate it.

evan-cz (approved):

evan-cz (approved):

PR #141, opened 2025-04-09
Initial Pulumi IaC work

spkane:
Initial Terraform work. It might make sense to move all ot some of this into another repo, but that can be discussed a bit later.


terraform/aws-eks-cluster/.terraform.lock.hcl
josephbarnett, line 1: Is this lock file needed or should it be part of a `.gitignore`?

spkane: It can be handled either way. It is similar to the `go.sum` file. Putting it in helps protect against supply chain attacks, requires a bit more maintenance. One can go either way with it.

terraform/aws-eks-cluster/k8s_cluster.tf
```diff
@@ -0,0 +1,48 @@
...
+  # Gives Terraform identity admin access to cluster which will
+  # allow deploying resources (Karpenter) into the cluster
+  enable_cluster_creator_admin_permissions = true
+  cluster_endpoint_public_access           = true
+
+  cluster_addons = {
+    coredns                = {}
+    eks-pod-identity-agent = {}
+    kube-proxy             = {}
+    vpc-cni                = {}
```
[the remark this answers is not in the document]
spkane: It mostly depends on how many resources the pods use. I mean we can fill them with one pod, or a ton of small pods. And all this are things I assume we'll adjust as we get closer to actually having tests to run. It could even adjust between some types of tests, if that is desirable in some cases.

spkane:
> Looks fine to me, I wonder about how to control pod density and if we need to do special configurations to ensure minimal ENI use (and nat addressing on a node).
> 
> Things I am talking about are:
> 
> * https://repost.aws/questions/QUQapIB9nwRCWlBAwmNqTKiA/change-max-pod-count-in-eks
> * https://docs.aws.amazon.com/eks/latest/best-practices/vpc-cni.html

This is all stuff that we'll tweak as we go along. If there are specific limits or whatever that we want to enforce under normal testing situations we should be able to do that easy enough. Variables for things we want to be able to adjust sometimes, and hard-coding, account rules, or AWS quota enforcement for things we want to be to hard limits.

PR #143, opened 2025-04-11
[helm] Expose prometheus service

jalev:
Operators of the Cloudzero Agent service may be interested to know what Prometheus is doing under the covers (e.g. storage, networking, etc) to help administrators of the service debug client-side things. One way to do that is to expose the underlying Prometheus service and let users roll their own monitoring.

I would have also made a ServiceMonitor object to go along with this (locked behind a flag) but I don't know if you want to commit the project to any one particular monitoring tool.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cloudzero-agent
 description: A chart for using Prometheus in agent mode to send cluster metrics to the CloudZero platform.
 type: application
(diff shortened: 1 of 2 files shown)


evan-cz:
Thanks again for this, @jalev! It was great talking to you on Zoom, and we /really/ appreciate the feedback.

As discussed, this would unfortunately conflict with some short-term plans we have so I'm going to close it, but please always feel free to reach out with stuff like this; we're happy to support stuff like this in general.

PR #144, opened 2025-04-11
CP-27884: fix agent-validator initialization

evan-cz:
Previously, we were relying on a bit of scripting to copy the agent-validator over to an emptyDir that could be used in the Prometheus image. However, since we are now basing everything off a scratch image that doesn't have a shell (or `cp` for that matter) installed that no longer works.

This adds an `install` command to the `cloudzero-agent-validator` image, which will copy itself to the provided destination, and alters the Helm chart to use that instead of `/bin/sh` and `cp`.

This also sneaks in a few other useful bits:

* Adds `package-debug` and `package-build-debug` make targets to build debug images (which do contain /bin/sh and cp)
* Adds a check to `scripts/ci-checks.sh` to make sure there is somewhere in the Helm chart that we know how to update for releases in order to ensure we don't break the release-to-main workflow.
* Rolls back the automated bump to 1.1.0-beta-3 so the workflow has something to do when we re-release it.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -127,7 +127,7 @@ agent:
   # The container image used for most CloudZero components.
   image:
     repository: ghcr.io/cloudzero/cloudzero-agent/cloudzero-agent
-    tag: 1.1.0-beta-3  # <- Software release corresponding to this chart version.
+    tag: 0.10.0  # <- Software release corresponding to this chart version.
     digest:
     pullPolicy: IfNotPresent
 
--- a/Makefile
+++ b/Makefile
@@ -232,6 +232,13 @@ test-smoke: ## Run the smoke tests
 
 # ----------- DOCKER IMAGE ------------
 
+DEBUG_IMAGE ?= busybox:stable-uclibc
+
+# Define targets for Docker image variants
+#
+#  $1: Target name (package, package-debug, package-build, package-build-debug)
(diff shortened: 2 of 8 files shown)


helm/values.yaml
```diff
@@ -127,7 +127,7 @@ agent:
   # The container image used for most CloudZero components.
   image:
     repository: ghcr.io/cloudzero/cloudzero-agent/cloudzero-agent
-    tag: 1.1.0-beta-3  # <- Software release corresponding to this chart version.
+    tag: 0.10.0  # <- Software release corresponding to this chart version.
```
roberthocking, line 130: just want to double check that this is in fact the right image

roberthocking (approved):
Looks Good. Just one question about the image tag. Feel free to merge when that's cleared up.

helm/values.yaml
evan-cz: Yeah, it's right, but it's worth explaining since it's definitely weird:

When run the release-to-main GitHub Action, part of the workflow will set that tag (using sed, so the comment is super important) to the version you're releasing. It does that by creating a commit, then pushing it to both develop and main.

I put that in there just because I'm a little worried about the workflow if there is no change (`git add` won't do anything, so there will be nothing for `git commit` to do, then `git push`, etc.); *something* might return a non-zero exit code. With this, it will set the value back to 1.1.0-beta-3 as part of the release and everything should 🤞 work as intended.

PR #152, opened 2025-04-15
CP-27962: webhook server and aggregator deploys roll on config change

dmepham:
testing:
1. asserted that changing a shared config (region) rolls both aggregator and webhook server pods
2. asserted that changing only an insightController setting only rolls the webhook-server pod
3. asserted that changing only an aggregator setting only rolls the aggregator pod
4. asserted that changing a non insightsController or aggregator config does not roll pods
5. asserted that additional pod annotations can still be added

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -3,7 +3,7 @@ name: cloudzero-agent
 description: A chart for using Prometheus in agent mode to send cluster metrics to the CloudZero platform.
 type: application
 version: 1.1.0-dev
-kubeVersion: ">= 1.23.0"
+kubeVersion: ">= 1.23.0-0"
 
 maintainers:
   - name: CloudZero
--- a/helm/templates/_cm_helpers.tpl
+++ b/helm/templates/_cm_helpers.tpl
@@ -7,34 +7,34 @@ region: {{ .Values.region }}
 cluster_name: {{ .Values.clusterName }}
 destination: {{ include "cloudzero-agent.metricsDestination" . }}
 logging:
-    level: {{ .Values.insightsController.server.logging.level }}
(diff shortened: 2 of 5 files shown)


helm/templates/_cm_helpers.tpl
```diff
@@ -0,0 +1,80 @@
+{{/*
```
dmepham, line 1: we've talked a bit about breaking the _helpers.tpl down into smaller chunks. this is an effort to start that

helm/templates/cm.yaml
```diff
@@ -252,37 +220,4 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   config.yml: |-
-    cloud_account_id: "{{ .Values.cloudAccountId }}"
```
dmepham, line 255: it might be kind of controversial to move these configs to a tpl file, but the nice thing is that we can be sure we're only rolling the deploys for configs that affect them

josephbarnett (approved):
This is really cool - I might recommend syncing with @evan-cz since he is making a number of changes in this area also

helm/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: cloudzero-agent
 description: A chart for using Prometheus in agent mode to send cluster metrics to the CloudZero platform.
 type: application
 version: 1.1.0-dev
-kubeVersion: ">= 1.23.0"
+kubeVersion: ">= 1.23.0-0"
```
dmepham, line 6: @spkane my tests failed because my server version is `v1.30.10-eks-bc803b4`, which I think it interpreted as a pre-release version? do you think this change is acceptable?

PR #155, opened 2025-04-15
Pulumi codebase for EKS cluster

spkane:
---

A few notes from @josephbarnett :

> I wonder about how to control pod density and if we need to do special configurations to ensure minimal ENI use (and nat addressing on a node).
> 
> Things I am talking about are:
> 
> - https://repost.aws/questions/QUQapIB9nwRCWlBAwmNqTKiA/change-max-pod-count-in-eks
> - https://docs.aws.amazon.com/eks/latest/best-practices/vpc-cni.html


spkane:
moved here: https://github.com/Cloudzero/ops-k8s-testing-infra/

PR #157, opened 2025-04-15
CP-28068 Document required API scopes for Kubernetes agent

beckilee:
This PR updates the README with the required API scopes for the Kubernetes agent.


helm/README.md
```diff
@@ -12,7 +12,12 @@ For the latest release, see [Releases](https://github.com/Cloudzero/cloudzero-ch
 
 - Kubernetes 1.23+
 - Helm 3+
-- A CloudZero API key
+- [A CloudZero API key](https://docs.cloudzero.com/reference/authorization#creating-a-new-api-key) with the following scopes:
+  - `container-metrics_v1:abandon`
+  - `container-metrics_v1:get-status`
+  - `container-metrics_v1:legacy`
+  - `container-metrics_v1:upload`
+  - `insights:read_insights`
```
dmepham, line 20: I think maybe we should hold off on adding this? I think ideally we don't want to be using this. I know that @josephbarnett had a conversation going with adam about it, but not sure where it landed.

dmepham (approved):
lgtm, but maybe worth removing the insights scope for the time being

helm/README.md
josephbarnett: @dmepham I recommended this change until we can solve this API issue with the new token support. This will fix the API validation check on the agent.

josephbarnett (approved):

PR #160, opened 2025-04-16
CP-28093: unconditionally enable The Gator.

evan-cz:
This removes the aggregator.enabled configuration option, and removes usage of it in checks, treating it as though it is always true.

Tested by checking that the output of a `helm template` before (w/ the value set to true) and after the change is identical.


josephbarnett (approved):
Send IT! 🚀

PR #166, opened 2025-04-17
CP-27353: move pkg/ and cmd/ into app/ and app/functions/

evan-cz:
There are two commits in this PR. The first just moves a lot of packages, but there are no other code changes. It does contain a few tweaks to the .gitignore (the `VERSION` line was matching `app/diagnostic/prom/version`, so I prefixed it and a few others with a /)

The second one drops pkg/build and modifies the code to use app/build instead. The major change is that I removed the application name from the build configuration, and the GetVersion function. Previously, it was hard- coded to cloudzero-insights-controller, which was wrong for *most* of the applications anyways.


josephbarnett (approved):
NICE! 👯‍♂️

PR #169, opened 2025-04-17
CP-28183: simple code cleanup changes"

josephbarnett:
# What
> See commit history for easier to identify changes. This PR simply deletes unused things, and moves some packages to utils or other appropriate location.

- **remove dead code**
- **move to correct location**
- **consolidate storage directories**
- **move otel instrumentation to logging dir**
- **move k8s client under utils**
- **remove duplicate k8s client code**
- **move telemetry helper for validator**
- **move paralization helper to utils**
- **move log to utils**
- **rename lock to filelock to match impl**

# Why

Clean code is strong code!

# How tested
1. Unit, Integration, and Smoke Tests


evan-cz (approved):
😍.

PR #180, opened 2025-04-24
CP-28160: add Helm template generation test

evan-cz:
This adds a target to the Makefile to automatically regenerate a manifest as part of the `generate` Make target. The output goes to tests/helm/template/manifest.yaml, which is checked into the repo.

The purpose of this change is to make sure that every change to the output of the Helm chart is intentional; if you alter the Helm chart without a corresponding alteration of the generated output, a status check should fail. Due to the fact that the manifest is included in the repo, the effect of the changes should be apparent in the commit, and will be easy to notice during a PR review if not sooner.

I also refactored the Helm portion of the Makefile a little bit to extract some flags into a common variable, and switched to using a GitHub Action to install protoc instead of using apt so we can reliably specify the version.

I tested this by making a change to the Helm chart without updating the generated manifest, and the status check failed.

### DIFF
--- a/tests/helm/template/overrides.yaml
+++ b/tests/helm/template/overrides.yaml
@@ -1,0 +1,6 @@
+cloudAccountId: "1234567890"
+clusterName: "my-cluster"
+region: "us-east-1"
+apiKey: "049a3c8b-7259-4ad3-931a-73596c647cf8"
+
+jobConfigID: "3e4c0c56-e932-4c5a-88f7-0dcaa0ff6b12"
--- a/.github/workflows/golang-ci.yml
+++ b/.github/workflows/golang-ci.yml
@@ -103,8 +103,10 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      - name: Install protoc
-        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          version: "29.3"
       - name: Install tools
         run: make install-tools
       - name: Generate code
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
-helm/templates/*.yaml
-helm/values.yaml
+/helm/templates/*.yaml
+/helm/values.yaml
+/tests/helm/template/manifest.yaml
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,9 @@ help: ## Show this help message
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n\nTargets:\n"} \
 		/^[a-zA-Z_-]+:.*##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
 
+.PHONY: FORCE
+FORCE:
+
 # ----------- CLEANUP ------------
 
 CLEANFILES ?= \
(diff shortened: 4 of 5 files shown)


tests/helm/template/overrides.yaml
jake-cloudzero, line 1: Think we accidentally added this file with an API Key...

jake-cloudzero, line 1: Do we want to include this file?

tests/helm/template/overrides.yaml
```diff
@@ -0,0 +1,6 @@
+cloudAccountId: "1234567890"
+clusterName: "my-cluster"
+region: "us-east-1"
+apiKey: "049a3c8b-7259-4ad3-931a-73596c647cf8"
```
josephbarnett, line 4: I assume this is a "fake key" - we add during generation which is total junk... CC: @jake-cloudzero

tests/helm/template/overrides.yaml
josephbarnett: I think it is a junk UUID... just to satisfy the secret needed in the chart. This fake UUID allows the generation of the same `tests/helm/template/manifest.yaml` I believe.

josephbarnett (approved):
Looks ok. I suspect I will be your first customer with the generic webhook changes

tests/helm/template/overrides.yaml
dmepham: it is probably worth switching to use `existingSecret` instead of `apiKey` anyway if we're just using it to template, just o avoid confusion

tests/helm/template/overrides.yaml
evan-cz: Yes, very much so. The target in the makefile is

```make
tests/helm/template/manifest.yaml: tests/helm/template/overrides.yaml helm-install-deps FORCE
	@$(HELM) template "$(HELM_TARGET)" ./helm -f $< > $@
```

`$<` in make syntax refers to the first dependency, so we basically do `helm template cz-agent ./helm -f tests/helm/template/overrides.yaml`, which will fail without that file.

tests/helm/template/overrides.yaml
evan-cz: > I assume this is a "fake key" - we use to only satisfy the field for comparison check ... total junk... CC: @jake-cloudzero

Correct. It's just `uuidgen | tr '[:upper:]' '[:lower:]'` (which is one of my very few aliases; `genuuid`).

Interestingly, the GitHub secrets scanner (sort of) [picked this up](https://github.com/Cloudzero/cloudzero-agent/security/secret-scanning/2), which was a pleasant surprise.

> it is probably worth switching to use existingSecret instead of apiKey anyway if we're just using it to template, just o avoid confusion

I don't think we should *switch*, but I do think it might be worthwhile to expand on this infrastructure to test different configurations, including using `existingSecret` instead of `apiKey`. It wouldn't be hard to just add a bunch of different override files and generated manifests, and just run `helm template` for each.

I probably should have added a comment to note that it is a fake key, though. If we do end up adding more tests like this I'll make sure to add one when I do that.

dmepham: yeah I think adding more configurations is a great idea. if we go down that route, it might be worth looking at something [like helm-unittest](https://github.com/helm-unittest/helm-unittest)

evan-cz: I haven't used helm-unittest (nor have I done this type of test before with Helm), but looking at it I'm not sure how I feel about it... it seems valuable, and I certainly wouldn't be opposed to adding some tests using it, but I think this method has some advantages.

First off, there is basically full coverage automatically. It seems like with helm-unittest you would have to add tests for the very specific things you want to check (e.g., "does this specific value in values.yaml match this specific value in the manifest?"), whereas with this what we're looking for is unexpected changes in *any* part of the chart. Basically, instead of having to write tests to verify that each little thing in the chart behaves as expected, with this we're more looking to make sure that whatever change we're making in the current commit has the desired effect, and no more.

I also think it's really cool that changes will show up in a very understandable way in PRs. You don't have to deeply understand a test, think about how the query maps to the underlying chart, etc. You just see a diff of the Helm chart and can pretty much tell at a glance what is going on.

I guess my point is that I don't think we should think of helm-unittest as an alternative to this; they complement each other fairly well. It seems like it could be a very cool supplement, though!

PR #181, opened 2025-04-24
CP-28409: enable building go binaries in parallel in Dockerfile

evan-cz:
## Why?

Build speed.

## What

Build binaries in parallel in Docker.

## How Tested

Works locally, I'm mostly curious about what CI makes of it.


josephbarnett (approved):

evan-cz:
Looks like this makes it ~ 3 minutes slower!

PR #185, opened 2025-04-24
Replicated test matrix workflow

spkane:
## Why?

> This awesome thing improves my smile brightness

- This tests a bunch of different types of Kubernetes cluster, including `eks`, `gke`, `aks`, and `rke2` (Rancher - which should be useful since people do run Rancher clusters in these cloud providers).
  -  Today, it is also possible to add support for `oks` (Oracle), `okd` (OpenShift) , `k3s`, and `kind`, but these do not really do us any good, so let's not waste the money.
- It also adds tests for `arm64` based systems, in addition to the default `amd64` based systems.
- And it also refactors the Docker build workflow to speed it up by using native runners for `arm64` images.

## What

A new github workflow.

## How Tested

Iteration...

### DIFF
--- a/.github/workflows/test-matrix-replicated-k8s-clusters.yaml
+++ b/.github/workflows/test-matrix-replicated-k8s-clusters.yaml
@@ -1,0 +1,403 @@
+name: test-matrix-replicated-k8s-clusters
+
+# NOTE: There is a `replicated` CLI should anyone be interested:
+# https://docs.replicated.com/reference/replicated-cli-installing
+
+on:
+  workflow_dispatch:
+    inputs:
+      image-repo:
+        description: "The registry for the image"
+        default: "ghcr.io"
+        type: "string"
+      image-path:
+        description: "The path for the image (no registry, no tag)"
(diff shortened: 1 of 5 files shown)


.github/workflows/test-matrix-replicated-k8s-clusters.yaml
```diff
@@ -0,0 +1,403 @@
...
+      contents: read
+      packages: read
+    outputs:
+      cluster-test-status: ${{ job.status }}
+    strategy:
+      fail-fast: false
+      matrix:
+        cluster:
+          [
+            {
```
josephbarnett, line 59: Curious - are these the active clusters configurations for replicated?

Why rancher?

spkane: These are the clusters that we are creating and testing with each test.

I left rke2 in, as I have known companies that use it for production in various cloud environments, so it seemed worth testing, although it certainly isn't a hard requirement.

PR #186, opened 2025-04-25
CP-27248: extend support for k8s objects (G/V/K variations) 

josephbarnett:
- **ensure obserability for cz_webhook_types_total metrics**
- **refactor hooks to use generics - and extend support for different types of k8s objects decoding**

## Why?

> This awesome thing improves my smile brightness

## What

## How Tested


tests/helm/template/manifest.yaml
```diff
@@ -577,6 +577,8 @@ data:
           match: exact
         - pattern: "storage_write_failure_total"
           match: exact
+        - pattern: "cz_webhook_types_total"
+          match: exact
```
josephbarnett, line 581: @dmepham @evan-cz 🎉 I think I added it right this time ;-)

app/domain/backfiller/backfiller.go
```diff
@@ -1,67 +1,218 @@
...
-// resource store, and configuration settings. The Start method begins the scraping process, iterating through
-// all namespaces and collecting data from the specified resources based on the provided filters.
-//
-// The package also includes helper functions such as writeResources and writeNodes to handle the listing and
-// storing of resources in a paginated manner, ensuring efficient data retrieval and storage.
-//
-// This package is valuable for organizations that need to keep track of their Kubernetes resources and ensure
-// that their inventory is always up-to-date. It provides a robust and flexible solution for scraping and storing
-// Kubernetes resource data.
+// Package backfiller provides functionality to backfill Kubernetes Resource objects, and if enabled invokes the webhook domain logic
```
josephbarnett, line 4: This file required a heavy refactor once the `app/domain/handler` logic implemented generics. 

Additionally changes made:
1. Process in parallel
2. Support all types (provided k8sclient provided access)
3. Improve ownership quality.

After looking at this, I believe it might be worth adding logic to await until the webhook is up, to ensure we get all things before it's start time. But that might be handled in the job definition? @evan-cz @dmepham ?

app/domain/webhook/handler/cron.go
```diff
@@ -1,81 +1,78 @@
 // SPDX-FileCopyrightText: Copyright (c) 2016-2024, CloudZero, Inc. or its affiliates. All Rights Reserved.
```
josephbarnett, line 1: This entire directory was refactored to use golang generics.

app/domain/webhook/webhook.go
```diff
@@ -116,55 +156,85 @@ func NewWebhookFactory(store types.ResourceStore, settings *config.Settings, clo
...
-
-	// Note: handlers beyond this point will not capture labels/annotations and will later be used to correlate resources
-	// to cloud resources (providerID) - to the pod using them.
-	wc.register(AppsGroup, Version1, "replicasets", handler.NewReplicaSetHandler(store, settings, clock))                                          // ✓ new
-	wc.register(CoreGroup, Version1, "services", handler.NewServiceHandler(store, settings, clock))                                                // ✓ new
-	wc.register(CoreGroup, Version1, "persistentvolumeclaims", handler.NewPersistentVolumeClaimHandler(store, settings, clock))                    // ✓ new
-	wc.register("networking.k8s.io", Version1, "ingresses", handler.NewIngressHandler(store, settings, clock))                                     // ✓ new
-	wc.register("apiextensions.k8s.io", Version1, "customresourcedefinitions", handler.NewCustomResourceDefinitionHandler(store, settings, clock)) // ✓ new
-	wc.register("gateway.networking.k8s.io", Version1, "gateways", handler.NewGatewayHandler(store, settings, clock))                              // ✓ new
+	// Apps API
```
josephbarnett, line 159: All the types 🤩

app/domain/backfiller/backfiller.go
dmepham: yeah, I agree that the job definition is the place to put that. I think it could just be as easy as an initcontainer that waits for the webhook svc to return 200

tests/helm/template/manifest.yaml
dmepham: maybe I'm missing something, but I don't know that I understand our naming structure for our custom metrics. `cloudzero_*` was originally used for metrics produced by the insights-controller, and then `czo_` is observability metrics from cloudzero-agent components. is is going to be confusing having a third cloudzero prefix? would it make sense to just have this be `cloudzero_webhook_types_total`

app/config/webhook/const.go
```diff
@@ -19,8 +19,9 @@ const (
 	Service                  ResourceType = 10
 	CustomResourceDefinition ResourceType = 11
 	ReplicaSet               ResourceType = 12
-	PersistentVolumeClaim    ResourceType = 13
-	Gateway                  ResourceType = 14
+	PersistentVolume         ResourceType = 13
+	PersistentVolumeClaim    ResourceType = 14
+	Gateway                  ResourceType = 15
 )
```
evan-cz, line 25: Maybe we could switch to using iota here?

```go
const (
	Unknown ResourceType = iota
	Deployment
	StatefulSet
	Pod
	Node
	Namespace
	Job
	CronJob
	DaemonSet
	Ingress
	Service
	CustomResourceDefinition
	ReplicaSet
	PersistentVolume
	PersistentVolumeClaim
	Gateway
)
```

Seems nicer to me...

evan-cz (approved):

tests/helm/template/manifest.yaml
evan-cz: Yeah, it should be `czo_`. I was just talking to Joe, he's adding some utility functions so we don't have to remember prefixes.

FWIW, I think `cloudzero_` should be for cost metrics, `czo_` for observability, regardless of where in the platform they're coming from. It's nice to be able to split them up in the gator as needed based on the prefix instead of adding a case for each metric...

app/domain/backfiller/backfiller.go
josephbarnett: Done!

app/config/webhook/const.go
josephbarnett: Done

tests/helm/template/manifest.yaml
josephbarnett: I renamed it to `czo_` since there was already filters for observability. I don't have a major preference but did like a short simple prefix for faster matching.

app/domain/diagnostic/webhook/check.go
```diff
@@ -77,59 +122,75 @@
...
 		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/json")
 
-	// disable TLS verification - we are trying to see if we can get to the API
-	// not validate the cert chain.
-	// #nosec G402: InsecureSkipVerify is set to true intentionally for testing purposes
 	client := &net.Client{Transport: &net.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // #nosec G402
```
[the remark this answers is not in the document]
josephbarnett: Thius is a test - ignore

PR #198, opened 2025-04-29
CP-28109: separate config for cost and o11y metrics rotation interval

evan-cz:
## What

This allows us to set the rotation interval independently for Cost and Observability metrics.

## Why

The latter are lower-priority and can probably wait longer, but we generally want to get the cost metrics shipped to the CloudZero server much more quickly.

## How Tested

`make test` has some tests which rely on the setting. I also deployed to a cluster to verify the Helm changes. Just to make extra sure the settings were being picked up, I also added some logging to the DiskStore (locally, not in the patch) to dump the interval being used and verified that it corresponded to the relevant setting.

### DIFF
--- a/app/storage/disk/disk.go
+++ b/app/storage/disk/disk.go
@@ -46,6 +46,14 @@ func WithContentIdentifier(identifier string) DiskStoreOpt {
 	}
 }
 
+func WithMaxInterval(interval time.Duration) DiskStoreOpt {
+	return func(d *DiskStore) error {
+		d.maxInterval = interval
+		d.ticker.Reset(interval)
+		return nil
+	}
+}
+
 // DiskStore is a data store intended to be backed by a disk. Currently, data is stored in Brotli-compressed JSON, but transcoded to Snappy-compressed Parquet
 type DiskStore struct {
 	dirPath           string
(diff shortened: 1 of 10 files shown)


app/storage/disk/disk.go
```diff
@@ -93,8 +104,8 @@ func NewDiskStore(settings config.Database, opts ...DiskStoreOpt) (*DiskStore, e
 		rowLimit:         settings.MaxRecords,
 		id:               uuid.New().String()[:8],
 		compressionLevel: settings.CompressionLevel,
-		maxInterval:      settings.MaxInterval,
-		ticker:           time.NewTicker(settings.MaxInterval),
+		maxInterval:      settings.CostMaxInterval,
+		ticker:           time.NewTicker(settings.CostMaxInterval),
```
evan-cz, line 108: I decided to default to using the cost setting, but these get overridden in practice a few lines later if someone passes in a `WithMaxInterval` (which we do in `main.go`).

helm/values.yaml
```diff
@@ -538,8 +538,10 @@ aggregator:
   database:
     # Max number of records per file. Use this to adjust file sizes uploaded to the server. The default value is good in most cases.
     maxRecords: 1500000
-    # Max interval to flush a metrics file. This is mostly useful for smaller clusters with little activity.
-    maxInterval: 10m
+    # Max interval to flush a cost metrics file. This is mostly useful for smaller clusters with little activity.
+    costMaxInterval: 10m
+    # Max interval to flush an observability metrics file. This is mostly useful for smaller clusters with little activity.
+    observabilityMaxInterval: 30m
```
evan-cz, line 544: Worth calling this out. Not sure what we want the default to be for o11y metrics. 30m seemed reasonable to me, but 🤷

dmepham (approved):

josephbarnett (approved):

PR #204, opened 2025-05-02
changed mkdir to mkdirall to fix create dir error

jake-cloudzero:
@roberthocking pointed out an error where in some cases where the shipper restarts (but the pod does NOT restart), then there would be an error for a directory already created in the emptyDir.

Changing `Mkdir` to `MkdirAll` fixes this. If the directory already exists then it will just return null. In addition, I did a quick check throughout to ensure we are not making this mistake anywhere else. This was the only occurrence, so it must have slipped through the cracks. potentially we should all K8s tests to restart individual containers within the pod and ensure we get no error logs.


evan-cz (approved):

josephbarnett (approved):
Only concern would be if the volume doesn't mount - ie. should this ever happen?

jake-cloudzero:
If the volume does not mount then we have bigger problems

PR #211, opened 2025-05-05
CP-28708: fix domain for webhook service

evan-cz:
## Why?

https://github.com/Cloudzero/cloudzero-charts/issues/200

In some situations, the webhook-configuration configmap would contain an incorrect value for the domain, causing the backfill job to fail due to an inability to resolve the (incorrect) DNS entry for the webhook service.

## What

This patch just tweaks the helper to generate the name using the same function which is used to generate the service name in the service resource.

## How Tested

The issue can be reproduced by installing to a namespaces different from the chart name (i.e., `helm install -n cza cz-agent ./helm`). The backfill job will fail before this patch, and works after it.

### DIFF
--- a/helm/templates/_cm_helpers.tpl
+++ b/helm/templates/_cm_helpers.tpl
@@ -27,7 +27,7 @@ certificate:
   cert: {{ .tls.mountPath }}/tls.crt
 server:
   namespace: {{ $namespace }}
-  domain: {{ $namespace }}-{{ .server.name }}-svc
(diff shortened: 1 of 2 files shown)


josephbarnett (approved):
interesting - why the checksum?

evan-cz:
> interesting - why the checksum?

Yeah, I was a little surprised by that at first too. The checksum is the SHA-256 of the output from `cloudzero-agent.insightsController.configuration`, and this change is in that template.

PR #214, opened 2025-05-07
CP-27295: add documentation for most stuff in values.yaml

evan-cz:
## Why?

Documentation helps customers, as well as us, understand what everything is for.

## What

I was also able to remove a few pieces that we no longer reference, but for the most part this is just about documenting as much stuff as I could.

## How Tested

N/A; documentation-only change.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -339,26 +339,42 @@ validator:
 
 server:
   name: server
+  # Container image configuration for the server.
+  #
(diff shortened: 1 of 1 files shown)


josephbarnett (approved):
Have @dmepham  also review

dmepham (approved):

jake-cloudzero (approved):

PR #223, opened 2025-05-12
 Disable matrix

spkane:
This is temporary to keep the merge Q from causing an issue until the underlying problem with the report action is fixed.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,15 +1,19 @@
-name: DockerBuild
+name: docker-build
 on:
   merge_group:
     types: [checks_requested]
   push:
-    branches:
-      - develop
-      - main
+    branches-ignore:
+      - "WIP-*"
+      - "wip-*"
     tags:
       - "*"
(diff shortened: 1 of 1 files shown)


evan-cz (approved):
generally LGTM, but...

Instead of commenting everything out, maybe just add a if: 1 != 1 or something to the jobs we want to skip?

Example: https://github.com/Cloudzero/cloudzero-agent/blob/develop/.github/workflows/mirror-chart.yml#L24

.github/workflows/docker-build.yml
```diff
@@ -1,15 +1,19 @@
...
 on:
   merge_group:
     types: [checks_requested]
   push:
-    branches:
-      - develop
-      - main
+    branches-ignore:
+      - "WIP-"
+      - "wip-"
```
evan-cz, line 8: I guess this does this do a substring search?

spkane: nope. I need an glob. Fixing.

PR #232, opened 2025-05-15
CP-27814: add support for federated mode

evan-cz:
## Why?

This should allow us to scale up to very large clusters.

## What

This new mode uses a DaemonSet to run Prometheus on each node, all pointed at the same Aggregator instance. There is still a Deployment of a Prometheus agent, which is used for scraping things like the Aggregator itself, KSM, the webook server, etc. The DaemonSet Prometheus is currently really only responsible for the cAdvisor job on each node.

## How Tested

You'll want to set `defaults.federation.enabled=true` to activate the code.

I've added a `tests/helm/templates/federated.yaml` so we can see check differences as we make changes going forward, but this is also a nice way to see the differences between federated and non-federated mode; `colordiff -u tests/helm/template/manifest.yaml tests/helm/template/federated.yaml`.

I've also tested deploying to a cluster and verifying that we do indeed have an instance of Prometheus per node, even while adding and removing nodes.

I've also verified that we are getting cAdvisor `container_*` metrics in the collector by waiting a bit, then downloading some metrics from `/cloudzero/data/uploaded` in the aggregator pod.


app/http/middleware/middleware.go
```diff
@@ -59,13 +60,19 @@ func LoggingMiddlewareWrapper(next http.Handler) http.Handler {
 		route := r.URL.Path
 		method := r.Method
```
josephbarnett, line 62: Cool - That was annoying before

evan-cz: I moved that over to #239. With some additional improvements.

helm/templates/_helpers.tpl
```diff
@@ -579,6 +579,13 @@ Otherwise, it will be the CloudZero API endpoint.
 'http://{{ include "cloudzero-agent.aggregator.name" . }}.{{ .Release.Namespace }}.svc.cluster.local/collector'
 {{- end -}}
 
+{{- define "cloudzero-agent.maybeGenerateSection" -}}
```
josephbarnett, line 582: Wicked cool

helm/templates/_helpers.tpl
```diff
@@ -723,3 +723,19 @@ spec:
       {{- .matchLabels | nindent 6 }}
 {{- end }}
 {{- end -}}
```
josephbarnett, line 725: This has been needed for some time - thank you

helm/templates/agent-daemonset-cm.yaml
```diff
@@ -0,0 +1,99 @@
...
+data:
+  prometheus.yml: |-
+    {{- if and (not .Values.defaults.federation.enabled) .Values.prometheusConfig.configOverride }}
+    {{ .Values.prometheusConfig.configOverride | nindent 4 }}
+    {{- else }}
+    global:
+      scrape_interval: {{ .Values.prometheusConfig.globalScrapeInterval }}
+
+    scrape_configs:
+      {{- if .Values.prometheusConfig.scrapeJobs.cadvisor.enabled }}
```
josephbarnett, line 19: Would we ever not want the cadvisor scrape job? That is the fundamental source for consumption metrics.

helm/templates/agent-daemonset-cm.yaml
```diff
@@ -0,0 +1,99 @@
...
+          action: keep
+        kubernetes_sd_configs:
+        - role: node
+          kubeconfig_file: ""
+          follow_redirects: true
+          enable_http2: true
+      {{- end }}{{/* End cadvisor scrape job */}}
+
+      {{- if .Values.prometheusConfig.scrapeJobs.prometheus.enabled }}
+      - job_name: static-prometheus
```
josephbarnett: very interesting... 🤔 💅🏻 polished 🙌🏻

helm/templates/agent-daemonset-cm.yaml
evan-cz: I can't think of a reason. I'm sort of just reusing as much logic as possible from `agent-cm.yaml` here. I'm cool with removing it, but I think we should remove it in both places...

helm/templates/agent-daemonset-cm.yaml
```diff
@@ -0,0 +1,99 @@
...
+          - source_labels: [__name__]
+            regex: "^({{ join "|" (include "cloudzero-agent.defaults" . | fromYaml).prometheusMetrics }})$"
+            action: keep
+      {{- end }}
+    {{- end}}
+
+    remote_write:
+      - url: {{ include "cloudzero-agent.metricsDestination" . }}
+        authorization:
+          credentials_file: {{ include "cloudzero-agent.secretFileFullPath" . }}
```
josephbarnett: Silly question - some time ago we did this to push to the OLTP endpoint, since things are going to the gator - which doesn't require auth...do we still need it?

helm/templates/agent-daemonset.yaml
```diff
@@ -0,0 +1,189 @@
...
+      {{- include "cloudzero-agent.generateAnnotations" .Values.server.podAnnotations | nindent 8 }}
+      labels:
+        {{- include "cloudzero-agent.server.labels" . | nindent 8 }}
+        {{- if .Values.server.podLabels}}
+        {{ toYaml .Values.server.podLabels | nindent 8 }}
+        {{- end}}
+    spec:
+      {{- include "cloudzero-agent.generatePriorityClassName" (.Values.defaults.priorityClassName | default .Values.server.priorityClassName) | nindent 6 }}
+      serviceAccountName: {{ template "cloudzero-agent.serviceAccountName" . }}
+      initContainers:
```
josephbarnett, line 26: I'd say we only need the initContainer and lifecycle hooks runs of the validator on the non-daemonset mode. 

Otherwise we get multiple validations and registrations for the same cluster.

Screenshot 2025-05-16 at 15 47 59

helm/templates/agent-daemonset-cm.yaml
evan-cz: Yeah, I was noticing that while I was putting this together. I was just going to use CP-28161 to remove it everywhere... I'll make a note in that ticket to make sure I don't forget to look at this bit (in both agent-cm and agent-daemonset-cm.

helm/templates/agent-daemonset-cm.yaml
josephbarnett: I'd recommend it. @dmepham any objections?

helm/templates/agent-daemonset.yaml
evan-cz: Good point! I forgot about the registrations part; I was just thinking we might want to validate some more stuff in the future, like the prometheus.yml, but we don't want multiple registrations... I'll remove that.

josephbarnett (approved):
Looks great - one change I'd ask is remove the extra validator run (lifecycle comment) on the daemonset. That will cause havoc at our status endpoint.

helm/templates/agent-daemonset.yaml
evan-cz: ✅

helm/templates/agent-daemonset-cm.yaml
dmepham: hey sorry, I missed this. looks like a good change. the one thing I would say about being able to toggle scrape jobs is that it can be useful for debugging. kind of a niche situation, but there has been at least one case where I found it useful to turn off all but one scrape job to make sure that particular job was sending to the collector.

but, in that case, the better solution is probably some extra debug logs in gator anyway 🤷

tests/helm/template/federated-overrides.yml
```diff
@@ -0,0 +1,10 @@
+cloudAccountId: "1234567890"
+clusterName: "my-cluster"
+region: "us-east-1"
+apiKey: "049a3c8b-7259-4ad3-931a-73596c647cf8"
```
dmepham, line 4: I know this is a fake api key, but maybe in the future it might be a good idea to called it something like `fake-key` just to be super explicit

PR #240, opened 2025-05-19
CP-27814: remove unnecessary configurability

evan-cz:
## Why?

Since this is a new component, we don't need to retain backwards compatibility with all of the server values. This allows us to remove some complexity from the chart.

## What

There are also a few cleanups to remove vestigial properties which are no longer used, and a few minor cleanups for comprehensibility.

## How Tested

Deployed to a cluster, checked that I'm still getting metrics in the aggregator, in both federated and non-federated mode.

### DIFF
--- a/helm/templates/agent-daemonset.yaml
+++ b/helm/templates/agent-daemonset.yaml
@@ -33,7 +33,7 @@ spec:
           command:
             - /bin/sh
             - -c
-            - sed "s/\${NODE_NAME}/$NODE_NAME/g" /etc/config/prometheus/configmaps/prometheus.yml.in > /etc/config/prometheus/configmaps/processed/prometheus.yml
+            - sed "s/\${NODE_NAME}/$NODE_NAME/g" /etc/config/prometheus/configmaps/unprocessed/prometheus.yml.in > /etc/config/prometheus/configmaps/processed/prometheus.yml
           env:
             - name: NODE_NAME
               valueFrom:
@@ -41,7 +41,7 @@ spec:
                   fieldPath: spec.nodeName
           volumeMounts:
             - name: unprocessed-config-volume
-              mountPath: /etc/config/prometheus/configmaps/
+              mountPath: /etc/config/prometheus/configmaps/unprocessed/
             - name: processed-config-volume
               mountPath: /etc/config/prometheus/configmaps/processed/
       containers:
@@ -71,28 +71,12 @@ spec:
             {{- if .Values.server.env }}
             {{- toYaml .Values.server.env | indent 12}}
             {{- end }}
-          lifecycle:
-            postStart:
-              exec:
-                command:
-                  - /checks/cloudzero-agent-validator
-                  - diagnose
-                  - post-start
-                  - -f
-                  - /checks/app/config/validator.yml
-            preStop:
-              exec:
-                command:
-                  - /checks/cloudzero-agent-validator
-                  - diagnose
-                  - pre-stop
-                  - -f
-                  - /checks/app/config/validator.yml
           args:
-            {{ toYaml .Values.server.args | nindent 12}}
-            {{- if .Values.server.agentMode }}
+            - --config.file=/etc/config/prometheus.yml
+            - --web.enable-lifecycle
+            - --web.console.libraries=/etc/prometheus/console_libraries
+            - --web.console.templates=/etc/prometheus/consoles
             - --enable-feature=agent
-            {{- end }}
           ports:
             - containerPort: 9090
           readinessProbe:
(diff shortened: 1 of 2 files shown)


helm/templates/agent-daemonset.yaml
```diff
@@ -71,65 +71,46 @@ spec:
...
-            initialDelaySeconds: {{ default 30 .Values.server.readinessProbeInitialDelay }}
-            periodSeconds: {{ default 5 .Values.server.readinessProbePeriodSeconds }}
-            timeoutSeconds: {{ default 4 .Values.server.readinessProbeTimeout }}
-            failureThreshold: {{ default 3 .Values.server.readinessProbeFailureThreshold }}
-            successThreshold: {{ default 1 .Values.server.readinessProbeSuccessThreshold }}
+            initialDelaySeconds: 30
+            periodSeconds: 5
+            timeoutSeconds: 4
+            failureThreshold: 3
+            successThreshold: 1
```
dmepham, line 91: I'm not super opposed to removing these, but I think it's worth noting that being able to configure these things can be helpful for debugging. but, I've never had to do that for this chart specifically, so 🤷

helm/templates/agent-daemonset.yaml
```diff
@@ -71,65 +71,46 @@ spec:
...
-                  - -f
-                  - /checks/app/config/validator.yml
           args:
-            {{ toYaml .Values.server.args | nindent 12}}
-            {{- if .Values.server.agentMode }}
+            - --config.file=/etc/config/prometheus.yml
+            - --web.enable-lifecycle
+            - --web.console.libraries=/etc/prometheus/console_libraries
+            - --web.console.templates=/etc/prometheus/consoles
             - --enable-feature=agent
```
dmepham, line 79: does this mean the use can't set agent mode for the agent in daemonset mode? I think it could still be helpful for debugging, no?

helm/templates/agent-daemonset.yaml
```diff
@@ -71,65 +71,46 @@ spec:
...
+              subPath: prometheus.yml
               readOnly: true
             - name: cloudzero-agent-storage-volume
-              mountPath: {{ .Values.server.persistentVolume.mountPath }}
-              subPath: "{{ .Values.server.persistentVolume.subPath }}"
-            - name: lifecycle-volume
-              mountPath: /checks/
-            - name: validator-config-volume
-              mountPath: /checks/app/config/
+              mountPath: /data
```
dmepham, line 112: I may be misremembering, but one thing that has been noted about the chart is that we mount to `/data`, but prometheus writes to `/prometheus/data` by default. we could either mount `/prometheus`, or reference `/data` in the prometheus args to fix this. the change is not that big a deal either way, and isn't really related to your change, so feel free to ignore

but just wanted to note that changing this to `/prometheus` might be a nice little improvement

helm/templates/agent-daemonset.yaml
```diff
@@ -71,65 +71,46 @@ spec:
             {{- if .Values.server.env }}
             {{- toYaml .Values.server.env | indent 12}}
             {{- end }}
-          lifecycle:
```
dmepham, line 74: I may be missing something, but why don't we want the daemonset agent pods to run these? is it because we don't want them all posting status from different places? that does make sense. maybe would be nice in the future to let the pod startup fail but not post to status

dmepham (approved):

helm/templates/agent-daemonset.yaml
evan-cz: Correct. We still have the tests running in the Prometheus from the Deployment, running it on each node too would create problems. I'm definitely open to adding checks that run on each node in the future... more tests is pretty much always better IMHO :)

dmepham: ah gotcha, forgot that we'd still have it on the deployment. thanks!

helm/templates/agent-daemonset.yaml
evan-cz: Not without tweaking the chart. Which, honestly, I think should be okay. AFAIK disabling agent mode would be useful for debugging, but would also effectively break the chart, right? In this case I'd rather not expose that option to users.

If we do want to expose it, we should think carefully about how; maybe add a section containing debugging variables that are very very clearly marked with text explaining that they are for debugging purposes only, and if you leave them on you're likely to end up with a broken chart or, at best, poor performance.

If it *doesn't* break the chart, I wouldn't mind just having a big toggle to set debugging = true on the chart, which would do this, crank up log levels, etc...

helm/templates/agent-daemonset.yaml
evan-cz: Oh, interesting, I wasn't aware of that. Yeah, I agree that changing it to /prometheus is a great idea for a follow up; I'd like to minimize the delta from upstream...

PR #241, opened 2025-05-19
CP-29142: complete Helm schema validation.

evan-cz:
## What

This adds a lot of additional validation to the values.yaml schema, as well as some supporting infrastructure. Specifically, lots of tests and the schema is now written in YAML, which is automatically processed to JSON (as Helm expects) in the Makefile.

There is also a script to merge the JSON Schema for Kubernetes (see https://github.com/yannh/kubernetes-json-schema) into our schema, so we can reference the Kubernetes types instead of redefining a ton of stuff. They're not as strict as I'd like, but they are very complete as they are auto-generated from Kubernetes. In other words, virtually zero false positives, but false negatives are possible.

This also add several missing properties to values.yaml and schema. These properties are referenced in the Helm chart, but were not previously mentioned in the values.yaml, nor included in the schema.

Oh, and I added a helm/validations.yaml to check some things we can't really validate in the schema. For example, the only test in there currently is one to ensure that either an `apiKey` or `existingSecretName` is set.

## Why?

This provides a couple of big advantages. First, it documents the format of a lot of stuff in the Helm chart, which was honestly a bit tricky to figure out.

Because Helm will perform validation against the schema by default, it also allows us to catch many configuration errors much sooner. In particular, because we use `additionalProperties: false` pretty much everywhere, it protects against users providing configuration which isn't actually used anywhere, but that they expect to be.

This also provides a significantly improved upgrade path for customers who are currently relying on deprecated, unstable APIs... at some point in the future, we can simply remove the deprecated items from the schema and customers will get an error message instead of their configuration overrides just silently no longer working.

I also have some ideas for things we could do with this moving forward. For example, Helm doesn't currently warn if you use a property marked as deprecated in the schema, but I believe it would be possible to use jq/gojq to extract a list of deprecated properties from the schema and create a Helm template to compare them to their default values and, if they don't match, emit a warning (or error, depending on the config). That way we can mark properties as deprecated and give customers time to update before we remove them entirely.

## How Tested

At its core, testing is pretty straightforward; just do what you would normally do with the chart! Helm will run the validations when you use the template, lint, install, etc. commands.

Beyond that, I also added a bunch of tests to the tests/helm/schema/ directory, as well as the code to run those tests to the Makefile (`make helm-test-schema`). This are all pretty simple, and can be either positive or negative tests based on their extension (*.pass.yaml vs *.fail.yaml).

### DIFF
--- a/helm/schema/k8s.json
+++ b/helm/schema/k8s.json
@@ -1,88 +1,5 @@
 {
-  "$defs": {
-    "com.cloudzero.agent.dns": {
-      "additionalProperties": false,
-      "description": "Specifies DNS configurations, both the dnsPolicy and dnsConfig.\nFor documentation, see https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config\n",
-      "properties": {
-        "config": {
-          "$ref": "#/$defs/io.k8s.api.core.v1.PodDNSConfig",
-          "description": "DNS configuration to use.\nFor documentation, see https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config\n"
-        },
-        "policy": {
-          "description": "DNS policy to use.\nFor documentation, see https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy\n",
-          "oneOf": [
-            {
-              "$ref": "#/$defs/io.k8s.api.core.v1.PodSpec/properties/dnsPolicy"
-            },
-            {
-              "type": "null"
-            }
-          ]
-        }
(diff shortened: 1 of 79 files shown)


helm/schema/k8s.json
evan-cz, line 1: Note that this file is just a copy of https://github.com/yannh/kubernetes-json-schema/blob/master/master/_definitions.json

It is merged with values.schema.yaml to produce values.schema.json, so we can reference k8s types in the schema instead of reinventing the wheel (pun about the k8s logo fully intended).

josephbarnett (approved):

PR #247, opened 2025-05-21
CP-29248: clean up and document Prometheus config

evan-cz:
## Why?

When looking at a recent customer issue, I thought there was a possibility that it was related to the Prometheus configuration, so I started looking at the configuration to understand it. As part of the process of figuring out how it worked, I decided to just document it in-place.

Once I was doing that, I noticed that there were also a lot of properties we were setting to the same values as the defaults, which was just creating a lot of extra complexity and cognitive load.

## What

This adds a lot of documentation to the configuration, which honestly should make things a lot easier on us (well, at least me) in the future when we need to work on the configuration. It also removes properties which needn't be set as they are already the defaults.

## How Tested

Deploying and verifying that we're still getting all the metrics was the big testing thing.

However, It's also useful to note that I tried to do this in three stages, with one commit each, to make reviewing substantially easier.

1. Add some indentation. There were no "real" changes in this commit, I just indented some list items a bit more to prepare it for some documentation.
2. Add documentation. Again, no actual changes; just adding a bunch of comments.
3. Remove superfluous properties. I went through the various properties, and checked the [Prometheus config documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/) to see if we actually needed the property, and removed the ones where we were just setting the value to the default.

Therefore, I would strongly recommend reviewing this commit-by-commit. It looks big and nasty when you view everything together, but when you look at each commit individually things will look a *lot* cleaner and you'll see that I didn't really do anything.

### DIFF
--- a/helm/templates/_cm_helpers.tpl
+++ b/helm/templates/_cm_helpers.tpl
@@ -136,6 +136,12 @@ remote_write:
 
 {{/* Define static-kube-state-metrics scrape job configuration */}}
 {{- define "cloudzero-agent.prometheus.scrapeKubeStateMetrics" -}}
+# Kube State Metrics Scrape Job
+# static-kube-state-metrics
+#
+# Kube State Metrics provides the CloudZero Agent with information
+# regarding the configuration and state of various Kubernetes objects
+# (nodes, pods, etc.), including where they are located in the cluster.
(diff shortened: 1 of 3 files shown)


dmepham (approved):
nice, thanks for the improvements and the very clear PR guide!

PR #248, opened 2025-05-22
CP-27861: additional checks for validator

jake-cloudzero:
This is a set of changes to the validator and helm chart to better detect config changes and report those upstream to help Cloudzero debug remote user issues more effectively.

This change includes adding a new Job to the Helm chart which will report on more information about the cluster, and send over configuration snapshots.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -725,7 +725,7 @@ aggregator:
   database:
     # Max number of records per file. Use this to adjust file sizes uploaded to
     # the server. The default value should generally be left unchanged.
-    maxRecords: 1.5e+06
+    maxRecords: 1500000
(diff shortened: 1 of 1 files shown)


evan-cz (approved):
LGTM. Love that you didn't have to add anything to the values.yaml!

helm/values.yaml
```diff
@@ -725,7 +725,7 @@ aggregator:
   database:
     # Max number of records per file. Use this to adjust file sizes uploaded to
     # the server. The default value should generally be left unchanged.
-    maxRecords: 1500000
+    maxRecords: 1.5e+06
```
evan-cz, line 728: Ohhh yeah please revert this.

PR #249, opened 2025-05-22
CP-29278: add Helm to tools.go

evan-cz:
## Why?

Helm 3.18.0 includes a regression in the way values are written, which is causing our generated manifests to include an unwanted change when regenerated from with 3.18.0 (i.e., our laptops if we did a brew upgrade since 3.18.0).

## What

This just adds Helm to tools.go, so `make install-tools` should download a good version (3.17.3, currently).

## How Tested

`make install-tools && make helm-generate-tests && git diff`


jake-cloudzero (approved):
Works as expected on my machine!

PR #251, opened 2025-05-26
Cp 29336 shipper hardening

jake-cloudzero:
The shipper component needs to operate under the presumption that errors will happen. Avoiding crashes and recovering from errors when they happen is paramount in this app. This PR aims to harden the shipper against outside errors.

This turned out to be a complete re-write of most of the shipper client code. What started as gentle optimizations resulted in an entire new house being built. The entire processing loop of the shipper is much more simple, with most branch paths now being tested under either Unit tests or Smoke tests. All in all, this gives us a much more durable aggregator deployment.

Changes:
- Change the base HTTP client to use the https://github.com/hashicorp/go-retryablehttp library
- Replay request logic has now been merged into the main file processing loop. The response we get from the server gives us all the information we need to do this (including valid URLs)
- More smoke tests testing error scenarios. There is still work to be done here, but this is closer to what we are looking for
- Increases the interval and timeout for the shipper cycle. It was at 1m in some configurations, this is too fast.
- Complete re-write of the disk monitoring code. The original implementation was quite inefficient, and contributed to the memory problem. In addition, this now accounts for in-memory filesystems as well, and is more greedy when reclaiming space for memory FS.
- Additional parameters added in the values config file to cover logging and http retries.
- Added an optional debug container that can be deployed along side the aggregator to allow for exploration of the filesystem mounts.


app/domain/shipper/shipper.go
```diff
@@ -262,40 +257,111 @@ func (m *MetricShipper) HandleRequest(ctx context.Context, files []types.File) e
...
+
+			// parse the new files with their presigned urls
 			for _, file := range chunk {
+				if purl, exists := urlResponse.Allocation[GetRemoteFileID(file)]; exists {
+					requests = append(requests, &UploadFileRequest{
+						File:         file,
+						PresignedURL: purl,
+					})
+				} else {
+					abandonRequests = append(abandonRequests, &AbandonAPIPayloadFile{
```
josephbarnett, line 286: I don't think we want to abandon a file if we didn't get a url. We should skip it and do it again later.

josephbarnett: Abandon should only be used if we can't honor a "replay" request. ie. file no longer exists. We don't want to use it for any other reason.

mock/remotewrite/pkg/remotewrite.go
```diff
@@ -44,11 +50,13 @@ type file struct {
...
-	S3AccessKey  string
-	S3PrivateKey string
-	UploadDelay  time.Duration
+	APIKey               string
+	S3Endpoint           string
+	S3AccessKey          string
+	S3PrivateKey         string
+	UploadDelay          time.Duration
+	ReplayRequestPayload string
+	ErrorOnUpload        bool
```
josephbarnett, line 59: Is this intentionally part of the change?

app/domain/shipper/http.go
```diff
@@ -7,44 +7,234 @@ import (
...
+)
+
+// HTTPClient interface for dependency injection and testing
+type HTTPClient interface {
+	// Do sends the http request
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// SendHTTPRequest performs an HTTP requests with retry logic
+func SendHTTPRequest(
```
josephbarnett, line 35: This is a really big method. I wonder if we'd benefit from using a special lib for exponential backoff and retry such as:
eg. https://github.com/hashicorp/go-retryablehttp

app/domain/shipper/shipper.go
```diff
@@ -262,40 +257,111 @@ func (m *MetricShipper) HandleRequest(ctx context.Context, files []types.File) e
...
+								Reason:      "failed to create a file object from this path",
+							})
+						}
+					}
+				}
+
+				return nil
+			}); err != nil {
+				logger.Err(err).Msg("failed to walk the data directory")
+			}
```
josephbarnett, line 330: I wonder if this could be moved to disk:
> See: https://github.com/Cloudzero/cloudzero-agent/pull/262

app/domain/shipper/http.go
jake-cloudzero: great look. now we dont have to own code

josephbarnett (commented):
A few questions -- then I think this looks right.

josephbarnett (approved):
🚀 Send it. Great Job @jake-cloudzero

PR #252, opened 2025-05-27
CP-29254: add default PodDisruptionBudget

evan-cz:
## Why?

We have seen our pods getting preempted when under load, which can cause data loss. We want to make sure that there is always at least one instance of most of our components is running.

## What

Added a podDisruptionBudget section to the defaults in the configuration, then tweaked the code to use that for the defaults. This still allows us to override the PDB for for individual components if needed.

I also threw in a commit to make the checksum in the validator job use the jobConfigID if provided, just to reduce noise in the future when the configuration changes.

## How Tested

Deploy. FWIW all we're really doing here is tweaking the defaults and hooking into mechanisms that already exist for generating PDBs.

### DIFF
--- a/tests/helm/template/federated.yaml
+++ b/tests/helm/template/federated.yaml
@@ -2092,16 +2092,6 @@ spec:
         - name: cloudzero-api-key
           secret:
             secretName: cz-agent-api-key
-      
-      affinity:
-        podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - podAffinityTerm:
-              labelSelector:
-                matchLabels:
-                  app: webhook-server
-              topologyKey: kubernetes.io/hostname
-            weight: 100
 ---
 # Source: cloudzero-agent/templates/init-cert-job.yaml
 apiVersion: batch/v1
@@ -2131,15 +2121,7 @@ spec:
         checksum/config: 28bf2e5c1e7aeb797e9370f5e86cc970f9cfffbe5d86a6be37c0c08ffbf934ca
     spec:
       
-      affinity:
-        podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - podAffinityTerm:
-              labelSelector:
-                matchLabels:
-                  app: webhook-server
-              topologyKey: kubernetes.io/hostname
-            weight: 100
+      
       
       serviceAccountName: cz-agent-cloudzero-agent-webhook-server-init-cert
       restartPolicy: Never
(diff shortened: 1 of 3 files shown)


helm/templates/validator-job.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: {{ include "cloudzero-agent.validatorJobName" . }}
   namespace: {{ .Release.Namespace }}
   {{- include "cloudzero-agent.generateAnnotations" .Values.defaults.annotations | nindent 2 }}
-    checksum/values: {{ .Values | toYaml | sha256sum }}
+    checksum/values: {{ (.Values.jobConfigID | default (. | toYaml | sha256sum)) | default (.Values | toYaml | sha256sum) }}
```
dmepham, line 7: I'm not sure I understand jobConfigID? is this to allow people to stop the job from running?

tests/helm/template/federated.yaml
```diff
@@ -2030,16 +2092,6 @@ spec:
         - name: cloudzero-api-key
           secret:
             secretName: cz-agent-api-key
-      
-      affinity:
```
dmepham, line 2034: it seems like the above change is causing this to be removed? I think we'd want to keep this, no? we want pods to try to avoid scheduling on the same nodes for availability

evan-cz: I put this in a separate commit (2153b7319a58d3f7218fa5904b4be9244c5ad273), the commit message has a bit of context. Basically, we only wanted to set this in one place, but ended up merging it into the defaults (oops!).

helm/templates/validator-job.yaml
dmepham: ah gotcha, thanks, makes sense. hopefully people don't set this thinking it's used for argo or something. maybe we could have the `NOTES.txt` raise an error if someone sets this in an install, and not a template? though, lots of tools just to a template anyway

I guess we could also `| grep -v checksum/values` in the command that generates the manifests? but there are potential issues with that too

dmepham: or maybe we could change the name to something like `_testId` or something? just spitballing. I'm fine with the change, just throwing out ideas

evan-cz: Hmmm, yeah I'm not sure how much we could really do to people from setting it. Not including it in values.yaml means they would really have to dig into the template to find it, so it's *probably* okay as is. I actually have another commit ready to go in a follow-up which isolates it even further, so it's only used in one spot in `_helpers.tpl`. I'll add a not there explaining it in detail, and that you should basically never use it... Maybe I'll hide it in a command-line `--set-string` to get it out of the tests/helm/template/*-overrides.yml too...

PR #263, opened 2025-06-03
Cp 29410 cluster configs refactor

jake-cloudzero:
This moves the cluster config loading mechanism out of the validator and into its own application


app/functions/cluster-config/loader/command.go
```diff
@@ -0,0 +1,237 @@
...
+	// ship to the remote
+	if err := post(c.Context, settingsValidator, &cfg); err != nil {
+		log.Ctx(c.Context).Err(err).Msg("failed to ship the config to the remote")
+	}
+
+	// should not return an error in most cases
+	return nil
+}
+
+func post(
```
dmepham, line 173: does this partially duplicate the post [here](https://github.com/Cloudzero/cloudzero-agent/blob/develop/app/utils/telemetry/post.go#L37)? or, maybe it doesn't make sense to re-use given that we're adding a custom header?

app/types/clusterconfig/clusterconfig.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package status;
+option go_package  = "../clusterconfig";
+
+message ClusterConfig {
```
dmepham, line 6: one thing that we will want in the future is a way to fetch KSM version and fetch TLS certificate information. 

that's my mistake for not mentioning those things. but, we can add those in a follow up PR

helm/templates/validator-job.yaml
```diff
@@ -28,10 +28,23 @@ spec:
           env:
             {{- include "cloudzero-agent.validatorEnv" . | nindent 12 }}
           command:
-            - /app/cloudzero-agent-validator
```
dmepham, line 31: it might be worth renaming this job and the related named templates, etc, from `validator` to `cluster-config` or similar

helm/templates/validator-job.yaml
```diff
@@ -28,10 +28,23 @@ spec:
...
+            - --region
+            - {{ .Values.region }}
+            - --cluster-name
+            - {{ .Values.clusterName }}
+            - --deploy-name
+            - TODO
+            - --chart-version
+            - {{ .Chart.Version }}
+            - --agent-version
+            - {{ .Values.components.agent.image.tag }}
```
dmepham, line 44: does this resolve correctly if they've set the image another way? I think they could also set this in the `default` map, or in the legacy `server`. it may be worth using the named template that @evan-cz added for getting the image (see [here](https://github.com/Cloudzero/cloudzero-agent/blob/develop/helm/templates/agent-deploy.yaml#L84)), and then decomposing that to get the image. or, using the same fallback logic there to get the actual tag that is rendered. whichever is easier and results in less duplication sounds good to me

dmepham (commented):
lgtm! I have some comments, but overall looks great

helm/templates/validator-job.yaml
evan-cz: This is an interesting problem...

If they've set the image another way this is actually probably safe, because `.Values.components.agent.image.tag` will actually still contain the version we're looking for. However, if they have overridden `.Values.components.image.tag` for some reason it won't.

Instead of passing this in as a command-line parameter, maybe it would be better to just use the version information built into the executable? This seems right to me since we are talking about the version of the agent here, not the chart.

Another interesting option I hadn't considered before, but could be pretty cool, would be to compare the version number from the chart with the version from the agent executable. We'd have to allow for the executable to be `dev-*` or something, but I'd love to catch it eraly if someone is using, say, v1.0.0 of the container with v1.2.0 of the chart...

Either way, if we want a version number from the chart that we can depend on, I think the key would be *not* getting it from something the user can override... How about we add something to _helpers.tpl to return the version number, then tweak the release-to-main workflow ([here](https://github.com/Cloudzero/cloudzero-agent/blob/99662d7f3e6811dd194adc09642de8ed85d87d15/.github/workflows/release-to-main.yml#L36-L41)) to also update that value? If that sounds good I can throw together a quick patch for you...

jake-cloudzero: @evan-cz If you could please do! I am about to push an update

evan-cz: Sort of; if it's in `.Values` the user can modify it. But yeah, I'm talking about creating a helper so you can do something like `{{ include "cloudzero-agent.versionNumber" }}`, and get something reliable.

But yeah, you could definitely mount the values ConfigMap and extract the info from there. Which makes a lot of sense to me since eventually we want to send that up to the API anyways, so why not just get access to it now...

helm/templates/config-loader-job.yaml
```diff
@@ -28,10 +28,23 @@ spec:
...
+            - --cluster-name
+            - {{ .Values.clusterName }}
+            - --release-name
+            - {{ .Release.Name }}
+            - --chart-version
+            - {{ .Chart.Version }}
+            - --agent-version
+            - {{ .Values.components.agent.image.tag }}
+            - --values-b64
+            - {{ .Values | toYaml | b64enc }}
```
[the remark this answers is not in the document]
evan-cz: ? That configmap [already exists](https://github.com/Cloudzero/cloudzero-agent/blob/018ab03eac7a642c649c0f0120235a55f79fcb0b/tests/helm/template/manifest.yaml#L683-L1251). Or am I misinterpreting?

helm/templates/validator-job.yaml
evan-cz: Oh, and `{{ (include "cloudzero-agent.versionNumber" . | fromYaml).version }}` to use it, if that wasn't clear.

jake-cloudzero: I do not see this variable defined anywhere?

evan-cz: What variable?

PR #264, opened 2025-06-04
CP-29585: periodically close HTTP connections

evan-cz:
## Why?

When Prometheus sends data to the Collector, it uses the standard Go HTTP client library, which reuses the connection as an optimization.

Unfortunately, the round robin to distribute the load across the k8s service happens when the initial connection is made, and since in non-federated mode we really only receive data from Prometheus and the Webhook server, which means we just have two things that create an initial connection, then always send data over those connections.

Luckily, and somewhat surprisingly (at least to me), Prometheus seems to be using HTTP/1.1 to send the data, which means we can just add a Connection: close periodically to get it to close the connection and go through the load balancing again.

## What

This will randomly (according to a configurable probability) add a `Connection: close` header to the HTTP response from the metrics collector endpoint.

## How Tested

Deploy; instead of seeing most of the activity in the logs of a singe pod it should now be a bit more evenly distributed. Set the reconnectFrequency to a lower value if you want it to be more aggressive, especially useful in a small cluster.

### DIFF
--- a/app/handlers/remote_write.go
+++ b/app/handlers/remote_write.go
@@ -5,6 +5,7 @@ package handlers
 
 import (
 	"io"
+	"math/rand"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
@@ -83,6 +84,18 @@ func (a *RemoteWriteAPI) PostMetrics(w http.ResponseWriter, r *http.Request) {
 		log.Ctx(ctx).Err(err).Msg("failed to put metrics")
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
+	}
+
+	// If we're using HTTP/1.x, we want to periodically close the connection to
+	// help distribute the load across the various collector replicas.
+	//
+	// Unfortunately this won't work for HTTP/2, but currently all traffic we're
+	// seeing from Prometheus is HTTP/1.1.
+	if r.ProtoMajor == 1 {
+		rf := a.metrics.Settings().Server.ReconnectFrequency
+		if rf > 0 && rand.Intn(rf) == 0 { //nolint:gosec // a weak PRNG is fine here
+			w.Header().Set("Connection", "close")
+		}
 	}
 
 	if stats != nil {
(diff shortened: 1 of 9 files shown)


app/handlers/remote_write.go
```diff
@@ -85,6 +86,18 @@ func (a *RemoteWriteAPI) PostMetrics(w http.ResponseWriter, r *http.Request) {
...
+	// If we're using HTTP/1.x, we want to periodically close the connection to
+	// help distribute the load across the various collector replicas.
+	//
+	// Unfortunately this won't work for HTTP/2, but currently all traffic we're
+	// seeing from Prometheus is HTTP/1.1.
+	if r.ProtoMajor == 1 {
+		rf := a.metrics.Settings().Server.ReconnectFrequency
+		if rf > 0 && rand.Intn(rf) == 0 { //nolint:gosec // a weak PRNG is fine here
+			w.Header().Set("Connection", "close")
+		}
```
josephbarnett, line 98: This causes a random reconnect - is there any way to make it more consistent? ie. some sort of counter, which we incr then div by reset factor? ie. ever 5th connection... ?

evan-cz: That would probably be possible, but there are some complexities which make it (IMHO) not worth it.

For one thing, if we did it we would probably want it to be per request per connection... Let's say we do it every 5th request; if we get 4 from Prometheus, then one from a webhook server, we would probably don't want a disconnect there, we would want to wait until the next one from Prometheus.

That means we'd need some connection tracking on the HTTP connections. Not only does that add overhead, it's pretty hacky; the Go APIs don't really make this easy. The only way I've found to do it (which is what I was doing to debug this) is to add middleware that tries to identify the connection based mostly on the remote IP and port.

If we tweaked go-obvious to add the connection tracking there, where we have access to the HTTP server object, I think this could be done a lot more reliably, but then we'd also have to find a way to expose that back to the handler. Possible, I just don't think it's worth the complexity.

FWIW, I actually really like stochastic solutions for problems like this. They tend to throw wrenches in the gears of fate, making the solution a bit more robust against unexpected snafus.

josephbarnett (approved):

PR #265, opened 2025-06-04
CP-29055: add support for setting Prometheus out_of_order_time_window

evan-cz:
## Why?

We are occasionally seeing errors from Prometheus: "Error on ingesting out-of-order samples". We haven't had much luck reproducing the issue, but based on some searches we're hoping this provides a way to get rid of them.

## What

This provides an easy way to configure the `out_of_order_time_window` property for Prometheus, and sets it to 5m by default.

## How Tested

Install the Helm chart to make sure it doesn't break anything. Whether it fixes the issue or not, we're not sure yet. Hence a draft PR for now.


josephbarnett (approved):
Pending testing - but looks good

PR #275, opened 2025-06-06
CP-29674: validate cloudAccountId contents in JSON Schema

evan-cz:
## Why?

This should catch some of invalid account IDs before deployment.

## What

Add a regex to the JSON Schema to validate the cloudAccountId field.

## How Tested

Manually :(. I tried to add tests to tests/helm/schema/, but in order to keep the test cases minimal we pass a `--set cloudAccountId=...` parameter (along with clusterName, region, and apiKey) when running the tests, so they get overridden anyways.


helm/values.schema.yaml
```diff
@@ -105,6 +105,17 @@ properties:
   cloudAccountId:
     type: string
     minLength: 1
+    # There are a few different possibilities, depending on the CSP:
+    #
+    # - AWS: Account ID is a 12-digit numeric identifier (possibly starting with a 0).
+    # - Google Cloud: this can be either a project name or project ID. IDs are
```
dmepham, line 111: I'd have to double check, but I think for GCP this needs to be project ID, and not project name, right @roberthocking ?

evan-cz: Hmmmm that would be surprising. They're used pretty interchangeably in the Google APIs...

dmepham: yeah I may be totally wrong, just have vague memories of having some distinction

evan-cz: Okay, I modified the JSON schema to use the project number. I also modified the scout to return the project number instead of the project ID....

I think we should be good to go now.

PR #288, opened 2025-06-16
CP-29735: allow 'global' top-level property, for subcharts

evan-cz:
## Why?

When the cloudzero-agent Helm chart is used as a subchart, Helm will automatically add a top-level 'global' property (for detais, see https://helm.sh/docs/chart_template_guide/subcharts_and_globals/).

What I didn't anticipate is that Helm will also include that property in the JSON Schema it is checking against, meaning that previously any use of the cloudzero-agent chart as a subchart would cause a schema validation error.

## What

This tweaks the JSON Schema to allow for a 'global' property, with any contents. It also adds a test which would have caught this issues, to ensure that there are no regressions.

## How Tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,7 +290,7 @@ HELM                      ?= helm
 
 HELM_ARGS = \
 	--namespace "$(HELM_TARGET_NAMESPACE)" \
-	--set cloudAccountId=$(CLOUD_ACCOUNT_ID) \
+	--set-string cloudAccountId=$(CLOUD_ACCOUNT_ID) \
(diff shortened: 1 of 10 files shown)


roberthocking (approved):

dmepham (commented):
just to clarify, we're not using `globals` in this chart, right? is it just for the KSM subchart? 
we've talked about pulling KSM out of a subchart and just use it directly, which would solve a lot of issues anyway

dmepham:
got it, I see, thanks!

PR #311, opened 2025-06-24
CP-29986: add additional properties to schema validation & values.yaml

evan-cz:
## Why?

These properties are used in the Helm chart, but were not present in values.yaml or the JSON Schema (which was based on values.yaml).

## What

This patch adds the properties to values.yaml, the JSON Schema, and adds a bunch of tests to validate many of the new properties.

I've also fixed a couple bugs in _helpers.tpl.

## How Tested

I added a bunch of tests. I also tested a bunch of stuff manually.


helm/values.yaml
```diff
@@ -12,6 +12,13 @@ apiKey: null
 # -- If set, the agent will use the API key in this Secret to authenticate with CloudZero.
 existingSecretName: null
 
+# -- Override the name of the chart. Used in resource naming.
+nameOverride:
```
dmepham, line 16: are these now included so that the validation passes? I know they're marked as deprecated. if they're needed here, does it make sense to leave a comment suggesting they not be used? `forceNamespace` especially is not so great, because we haven't been so consistent with using the named template that references it.
really we never should have had that one :/

dmepham (approved):

helm/values.yaml
dmepham: great thanks!

PR #314, opened 2025-06-25
CP-30020: add kubeconform tests

evan-cz:
## Why?

Previously, there was a bug in the template which was causing the generated output to not be valid according to the Kubernetes schema. I want to be able to catch bugs like this automatically.

## What

For every schema test where the result is valid (i.e., files named *.pass.yaml), run the output through kubeconform with strict validation enabled.

The schema tests are a great opportunity because they exercise a lot of different properties, particularly ones which don't get used very often. This gives us much greater coverage. In fact, I was able to spot several issues using these new tests, so this PR also includes fixes for those.

I also added a GitHub Action job to run `helm -j helm-test`. Oh, and that `-j` is interesting; I refactored the Makefile a bit to create multiple targets, so -j will actually cause these tests to be executed in parallel.

## How Tested

`make helm-install-deps install-tools && make -j helm-test`

Note that I'm currently targeting the CP-29985 branch. I'll update this PR to target develop once that is merged.

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -95,21 +95,37 @@ app.kubernetes.io/component: {{ .Values.server.name }}
 {{- end -}}
 
 {{/*
-Create unified labels for prometheus components
+Common base labels for all Kubernetes resources
 */}}
-{{- define "cloudzero-agent.common.metaLabels" -}}
-{{- $labels := dict
+{{- define "cloudzero-agent.baseLabels" -}}
+{{- dict
+    "app.kubernetes.io/name" (include "cloudzero-agent.name" .)
+    "app.kubernetes.io/instance" .Release.Name
     "app.kubernetes.io/version" .Chart.AppVersion
     "helm.sh/chart" (include "cloudzero-agent.chart" .)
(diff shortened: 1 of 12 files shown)


helm/templates/_helpers.tpl
```diff
@@ -642,13 +667,11 @@ Generate image configuration with defaults.
 {{- $tag         := (.image.tag         | default .defaults.tag) -}}
 {{- $repository  := (.image.repository  | default .defaults.repository) -}}
 {{- $pullPolicy  := (.image.pullPolicy  | default .defaults.pullPolicy) -}}
-{{- $pullSecrets := (.image.pullSecrets | default .defaults.pullSecrets) -}}
```
dmepham, line 645: why are we removing this? wouldn't we want it to be set in the same way still?

dmepham (approved):

helm/templates/_helpers.tpl
evan-cz: It's in the wrong spot. There is a `generateImagePullSecrets` invocation to generate it in the right spot.

dmepham: gotcha, thanks!

PR #315, opened 2025-06-25
CP-30042: allow resource_type and workload labels in Gator

evan-cz:
## Why?

Be the resource_type and workload labels, which are emitted by the webhook server.

## What

Add "resource_type" to the list of allowed labels.

## How Tested

YOLO.

### DIFF
--- a/helm/templates/_defaults.tpl
+++ b/helm/templates/_defaults.tpl
@@ -171,8 +171,10 @@ metricFilters:
         - product_name
         - provider_id
         - resource
+        - resource_type
         - unit
         - uid
+        - workload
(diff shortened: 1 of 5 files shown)


helm/templates/_defaults.tpl
```diff
@@ -171,6 +171,7 @@ metricFilters:
         - product_name
         - provider_id
         - resource
+        - resource_type
```
dmepham, line 174: I think we will want `workload` as well

dmepham (approved):

josephbarnett (approved):

PR #319, opened 2025-06-27
CP-30085: add cluster name detection to Google scout

evan-cz:
## Why?

While not all providers support detecting the cluster name automatically (I'm looking at you, Amazon EKS), Google Cloud does.  We may as well take advantage of the functionality where it exists.

## What

This adds support for detecting the cluster name on Google Cloud.

## How Tested

Remove the clusterName from your overrides, deploy to Google Cloud.

### DIFF
--- a/app/utils/scout/google/google.go
+++ b/app/utils/scout/google/google.go
@@ -18,9 +18,10 @@ import (
 
 const (
 	// GCP metadata service endpoints
-	metadataBaseURL = "http://metadata.google.internal/computeMetadata/v1"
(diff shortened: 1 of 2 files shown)


dmepham (approved):

evan-cz:
> One thing that would be cool is - what if the customer provides values, which don't match what is available via scout inspections? How can we do the right thing? ie. Warn and use the discovered value instead...

That's a really good question, and I don't have a good answer right now. It's certainly possible to warn and use the detected value. For now I treat supplied values as overrides, but I'm definitely open to other options.

PR #326, opened 2025-06-27
CP-30109: add automatic Istio sidecar injection suppression for webhook server

evan-cz:
## Why?

Previously, users deploying cloudzero-agent in Istio-enabled clusters had to manually configure `sidecar.istio.io/inject: "false"` annotations to prevent Istio sidecar injection from interfering with webhook TLS communication. This created a poor out-of-the-box experience and required users to read documentation and troubleshoot connectivity issues.

## What

This change makes the webhook server work automatically in Istio environments by adding the `sidecar.istio.io/inject: "false"` annotation by default. The webhook server pods now automatically include this annotation without any user configuration required.

The implementation adds a new
`insightsController.server.suppressIstioAnnotations` boolean property (default: false) that controls whether the automatic annotation is added. When false (default), the template logic automatically merges the Istio annotation with any user-provided `podAnnotations`. When set to true, users can disable this behavior and handle Istio configuration manually using the alternative approaches documented in `docs/istio.md`.

The annotation should have no effect in non-Istio environments.

## How Tested

Schema tests for the new property, plus the manifest generation tests look good, and for good measure I deployed it.

Haven't tested that istio does what we want, though; only that the annotation exists.

### DIFF
--- a/app/functions/helmless/default-values.yaml
+++ b/app/functions/helmless/default-values.yaml
@@ -668,6 +668,13 @@ insightsController:
     deploymentAnnotations: {}
     # Annotations to add to the webhook server pods.
     podAnnotations: {}
+    # Whether to suppress Istio-related annotations on webhook server pods. When
+    # false (default), the sidecar.istio.io/inject: "false" annotation is added
+    # to prevent Istio sidecar injection which can interfere with webhook TLS.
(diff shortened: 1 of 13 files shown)


dmepham (approved):
🥳

PR #329, opened 2025-07-01
CP-30077: add Azure scout

evan-cz:
## Why?

To automatically detect values in AKS

## What

Added an Azure scout.

The cluster name detection is a bit dodgy; the problem is that Azure IMDS provides a single string for both the resource group name and the cluster name: `MC_{resourceGroupName}_{clusterName}_{region}`.

Unfortunately, both resourceGroupName and clusterName can contain underscores, so without knowledge of either parsing is ambiguous. For example, for "MC_foo_bar_baz_qux_useast1", the clusterName could conceivably be bar_baz_qux, baz_qux, or qux.

While Kubernetes clusters *can* include underscores, it is considered best practice to avoid them (generally using dashes instead). Therefore, this scout will assume that the cluster name does *not* contain any underscores. If a cluster name contains underscores, you will need to manually specify the cluster name.

## How Tested

There are a bunch of tests included in this PR, but I've also tested by deploying to an AKS cluster.

### DIFF
--- a/app/config/validator/deployment.go
+++ b/app/config/validator/deployment.go
@@ -12,12 +12,20 @@ import (
 	"github.com/rs/zerolog/log"
 
 	"github.com/cloudzero/cloudzero-agent/app/utils/scout"
+	"github.com/cloudzero/cloudzero-agent/app/utils/scout/types"
 )
 
 type Deployment struct {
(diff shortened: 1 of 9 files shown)


jake-cloudzero (approved):
This stuff is pretty much in your domain... As long as the tests pass lets push on this

PR #332, opened 2025-07-02
CP-30251: Fix validator environment

jake-cloudzero:
## Why?

> The branch name is incorrect

There namespace aware values were not using the right namespace in the validator. We were looking for `NAMESPACE` when we should be looking for `K8S_NAMESPACE`. The confload job does it correctly, but we never updated the validator to use the shared values.

```
│ 2025-07-02T17:08:29Z │ 1.2.2         │ 1.26            │ {'k8s_namespace': {'message': 'the env variable `NAMESPACE` must exist',         │ POD_STARTED  │
│                      │               │                 │ 'passing': False}, 'k8s_provider': {'message': 'the env variable `NAMESPACE`     │              │
│                      │               │                 │ must exist', 'passing': False}, 'k8s_version': {'message': None, 'passing':      │              │
│                      │               │                 │ True}, 'kube_state_metrics_reachable': {'message': 'Failed to fetch metrics      │              │
│                      │               │                 │ after 12 retries', 'passing': False}, 'scrape_cfg': {'message': None, 'passing': │              │
│                      │               │                 │ True}, 'prometheus_version': {'message': None, 'passing': True},                 │              │
│                      │               │                 │ 'webhook_server_reachable': {'message': 'received non-2xx response from          │              │
│                      │               │                 │ https://cloudzero-server-webhook-svc.cloudzero.svc.cluster.local/validate after  │              │
│                      │               │                 │ 6 retries: webhook request failed: Post                                          │              │
│                      │               │                 │ "https://cloudzero-server-webhook-svc.cloudzero.svc.cluster.local/validate":     │              │
│                      │               │                 │ context deadline exceeded', 'passing': False}}
```

### DIFF
--- a/app/domain/diagnostic/k8s/namespace/check.go
+++ b/app/domain/diagnostic/k8s/namespace/check.go
@@ -7,10 +7,10 @@ package namespace
 import (
 	"context"
 	"net/http"
-	"os"
 
 	config "github.com/cloudzero/cloudzero-agent/app/config/validator"
 	"github.com/cloudzero/cloudzero-agent/app/domain/diagnostic"
+	"github.com/cloudzero/cloudzero-agent/app/domain/k8s"
 	logging "github.com/cloudzero/cloudzero-agent/app/logging/validator"
 	"github.com/cloudzero/cloudzero-agent/app/types/status"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


evan-cz (commented):
Does https://github.com/Cloudzero/cloudzero-agent/blob/4ef2385eb1e4364e29004fe8d1ca83bff3031014/app/config/webhook/server.go#L10 need to change, too?

jake-cloudzero:
No, we do not need it there because this is for the validator to be environmentally concious of where it is running, not for configs sake

evan-cz (approved):

PR #338, opened 2025-07-03
CP-29051: fix flaky tests

evan-cz:
## Why?

There are a few tests which occasionally fail in CI, wreaking havoc on my sanity.

## What

Fix (🤞) the tests.

## How Tested

By running them. They now work reliably locally, but honestly I wasn't able to reproduce them all so I'm not *certain* about the fixes.


jake-cloudzero (approved):
This will be nice

PR #347, opened 2025-07-11
CP-30492: optimize Docker build performance

evan-cz:
## Why?

Because building images was very slow.

## What

This represents a comprehensive overhaul of the Docker build process to achieve dramatically faster build times, going from 2:30-3:00 minutes down to ~12 seconds.

The optimization consists of several complementary strategies:

* Multi-stage build with platform-specific caching

  Each stage uses platform-specific cache mount IDs (e.g., id=apk-$TARGETPLATFORM, id=gomod-$TARGETPLATFORM) to ensure ARM64 and AMD64 builds don't interfere with each other's caches.

* Selective file copying instead of .dockerignore:

  Replaced the broad "COPY . ." approach with explicit copying of only the files needed for building (app/ and Makefile). This reduces the build context from ~656kB to ~37kB and eliminates the dependency on .dockerignore patterns.

* Conditional dependency generation:

  Added REGENERATE variable to Makefile that allows skipping the regeneration of app/functions/helmless/default-values.yaml during container builds. This eliminates the need to install Helm and Prettier tools inside the container, which was a significant performance bottleneck. The container build uses REGENERATE=never while normal development workflows continue to use the default REGENERATE=auto behavior.

  Setting this value in the Dockerfile allows us to avoid dependencies on Helm and Prettier. However, this may not be worthwhile in the medium-term, as it looks like the next version of Alpine (3.23) will have a 'prettier' package, then we can just add that & helm to the apk command.

## How Tested

`make package-build`. I did also test the resulting image by deploying the chart.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,0 @@
-# These are platform-specific, and can override using system-wide
-# versions of the tools.
-/.tools/bin
-/.tools/node_modules
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ REVISION       ?= $(shell git rev-parse HEAD)
 TAG            ?= dev-$(REVISION)
 OUTPUT_BIN_DIR ?= bin
 GIT_ROOT       ?= $(shell git rev-parse --show-toplevel)
+REGENERATE     ?= auto
 
 # Default Helm configuration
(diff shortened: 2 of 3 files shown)


jake-cloudzero (approved):
LETS GO

PR #368, opened 2025-07-23
CP-31009: add lots of detail to SECURITY.md

evan-cz:
## Why?

Previous version was pretty barren and unmaintained.

## What

Wrote a new security document.

## How Tested

PR reviews.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,14 +1,188 @@
 # Security Policy
 
-## Supported Versions
+## Overview
 
-Use this section to tell people about which versions of your project are
-currently being supported with security updates.
+CloudZero takes security seriously, especially for the CloudZero Agent, which is installed on customer infrastructure. We maintain a robust security posture through automated scanning, continuous monitoring, and proactive vulnerability management.
 
-| Version | Supported          |
-| ------- | ------------------ |
-| > 0.1.x | :white_check_mark: |
+## Security Commitment
 
-## Reporting a Vulnerability
+We are committed to:
 
-Please do not report security vulnerabilities on the public GitHub issue tracker. Email [security@cloudzero.com](mailto:security@cloudzero.com) instead.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -1,14 +1,188 @@
...
+#### Container Image Scanning
+
+All container images are automatically scanned for vulnerabilities using two complementary tools:
+
+- **[Trivy](https://trivy.dev/latest/)**: Comprehensive vulnerability scanner that detects OS and library vulnerabilities
+- **[Grype](https://github.com/anchore/grype)**: Fast vulnerability scanner with high accuracy
+
+**Scanning Configuration**:
+
+- **Severity Threshold**: High and Critical vulnerabilities trigger build failures
```
josephbarnett, line 31: First, let me say I ❤️ this version of the document. 

Element, which might be nice to add:
1. How long do we support a version for? I think we talked about 6 months
2. related to 1, how do we deliver fixes to CVEs which are deemed critical. Getting into version sprawl would be difficult to manage. Maybe we commit to delivering patches in a new release of the same version lineage (eg. minor version as to avoid major version breaking changes?)
3. Is there a policy document/material we need to commit to?
4. (maybe a different PR), but I believe we could generate SBOM files and add them to the build for easier scanning by clients.
5. (Def different PR) Do we have or need a signing requirement?
We should get @sean-calista-cloudzero and security to sign off also.

evan-cz: 1. Yes, please let's have a pre-determined support window. @teevans maybe you have some insight/opinions?
2. I don't hate the idea of doing security updates for stable branches. We're going to need to beef up our release process a bit, though.
3. Sounds like a question for Sean.
4. 🤔 I don't mind the idea, but I'm not sure about the audience... it's a cool optimization, but the people who care about security are probably already going to have the infrastructure in place to scan the container images.

josephbarnett: I say we "commit and iterate" - we can queue these points up for follow ups. Awesome work 🙌🏻

PR #371, opened 2025-07-24
fix: webhook config resource names to use correct plural forms

josephbarnett:
## Summary
- Fixed critical webhook configuration issue where resource names were using singular forms instead of required plural forms
- Updated all 18 resource names in `webhook-validating-config.yaml` to match Kubernetes API resource plurals
- This was preventing the webhook from properly intercepting operations on these resources

## Root Cause
The webhook configuration was using singular resource names (e.g., `deployment`, `pod`, `namespace`) but Kubernetes API resources are plural (e.g., `deployments`, `pods`, `namespaces`). This caused the webhook to fail intercepting operations on these resources.

**Issue introduced in:** https://github.com/Cloudzero/cloudzero-agent/pull/333

## Changes
- Updated all resource names in `helm/templates/webhook-validating-config.yaml` to use correct plural forms
- Verified against `kubectl api-resources --verbs=list -o wide --namespaced=false` output

## Test Plan
- [x] Verify webhook properly intercepts CREATE/UPDATE/DELETE operations on all configured resources
- [x] Test deployment of updated Helm chart
- [x] Confirm no regression in webhook functionality

**HOTFIX**: This is a critical fix for production webhook functionality.

### DIFF
--- a/tests/webhook/Makefile
+++ b/tests/webhook/Makefile
@@ -53,12 +53,9 @@ test-webhook-debug: check-deps
 
 help:
 	@echo "Available targets:"
-	@echo "  test-webhook               - Run basic webhook integration test (no API key)"
-	@echo "  test-webhook-debug         - Run basic test in debug mode (keep cluster)"
-	@echo "  test-webhook-chart         - Run Helm chart integration test (requires API key)"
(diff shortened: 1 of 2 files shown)


evan-cz (approved):

evan-cz:
I fixed the lint error (just had to regenerate some files), hope you don't mind.

PR #380, opened 2025-08-04
CP-28992: add Checkov to build system + CI, fix several issues

evan-cz:
## Why?

Checkov is a free security scanner which can scan our Helm chart to look for issues. Some of the issues it finds are a bit silly and we won't be "fixing", but others are good ideas and we should fix.

## What

This adds the initial Checkov support, and a file that ignores all the issues that are uncovered. Then there are several follow-up patches which fix several issues it uncovers.

## How Tested

`make analyze-checkov`

### DIFF
--- a/.checkov.yaml
+++ b/.checkov.yaml
@@ -7,10 +7,6 @@ skip-path:
 
 skip-check:
   # Rule:
-  #   Containers should have readiness probes to control traffic routing.
-  - CKV_K8S_9
-
-  # Rule:
   #   Containers should have CPU requests defined for proper scheduling.
   - CKV_K8S_10
 
--- a/helm/templates/agent-daemonset.yaml
+++ b/helm/templates/agent-daemonset.yaml
@@ -64,6 +64,16 @@ spec:
             timeoutSeconds: 10
             failureThreshold: 3
             successThreshold: 1
+          readinessProbe:
+            httpGet:
+              path: /healthz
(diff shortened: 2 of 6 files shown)


roberthocking (approved):

.checkov.yaml
```diff
@@ -0,0 +1,99 @@
...
+  #   Pods should have security context defined for proper isolation.
+  - CKV_K8S_29
+
+  # Rule:
+  #   Containers should have security context defined for proper isolation.
+  - CKV_K8S_30
+
+  # Rule:
+  #   Containers should use seccomp profiles for syscall filtering.
+  - CKV_K8S_31
```
josephbarnett, line 67: This would require we generate seccomp files, which is kind of PITA unless you are using ptrace. Not something I would support at this time.

evan-cz: I don't think we need to address all of these. I suspect there are several issues in here we don't actually want to "fix". In such cases, we can just add a note to the entry in this list explaining our decision.

The other possibility for ignoring issues, if we just want to ignore them in one spot not the whole chart, is to add an annotation to the resource. For example, there is one issue in there about pods not having certain capabilities when not needed (then one that runs all the kubectl commands)... in that case, I think it would make sense to just ignore the warning for that one resource, so we don't accidentally do it in other places where we don't actually need it.

PR #382, opened 2025-08-04
CP-29890: remove observability files on upload

jake-cloudzero:
## Why?

We are never going to need to replay observability files, so why store them in uploaded? We should only be using the customer's space for information they themselves will need.

## What

Quick change to the file processing code post-upload to remove observability files in addition to log files.

## How Tested

Our existing tests are enough to cover this change

### DIFF
--- a/app/domain/shipper/shipper.go
+++ b/app/domain/shipper/shipper.go
@@ -318,8 +318,8 @@ func (m *MetricShipper) HandleRequest(ctx context.Context, files []types.File) e
 						return err
 					}
 
-					// if log file, remove it. we do not need to store these after upload
-					if strings.HasPrefix(req.File.UniqueID(), disk.LogsContentIdentifider) {
+					// if not a metric file, remove it. we do not need to store these after upload
+					if !strings.HasPrefix(req.File.UniqueID(), disk.CostContentIdentifier) {
 						if err := os.Remove(req.File.Location()); err != nil {
 							logger.Err(err).Str("file", req.File.UniqueID()).Msg("failed to remove log file")
 						}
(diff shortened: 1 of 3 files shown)


app/domain/shipper/shipper.go
```diff
@@ -318,8 +318,8 @@ func (m *MetricShipper) HandleRequest(ctx context.Context, files []types.File) e
 						return err
 					}
 
-					// if log file, remove it. we do not need to store these after upload
-					if strings.HasPrefix(req.File.UniqueID(), disk.LogsContentIdentifider) {
+					// if log file (or observability file), remove it. we do not need to store these after upload
+					if strings.HasPrefix(req.File.UniqueID(), disk.LogsContentIdentifider) || strings.HasPrefix(req.File.UniqueID(), disk.ObservabilityContentIdentifier) {
```
evan-cz, line 322: I wonder if it might be prudent to reverse the logic here, and instead of looking for files with `disk.ObservabilityContentIdentifier` look for files _without_ `disk.CostContentIdentifier`? I suspect the only metrics we'll ever care about being able to replay are the cost metrics...

tests/smoke/client_test.go
```diff
@@ -77,7 +77,7 @@ func TestSmoke_ClientApplication_Runs(t *testing.T) {
 		// number of s3 files that are not logs
 		countUploaded := 0
 		for _, item := range response.Objects {
-			if !strings.HasPrefix(item.Key, "logs") {
+			if !strings.HasPrefix(item.Key, "logs") && !strings.HasPrefix(item.Key, "observability") {
```
evan-cz, line 80: Ditto

evan-cz (approved):
Great idea!

One small suggestion for the implementation, but once that's done LGTM!

PR #387, opened 2025-08-05
CP-31336: add CronJob for backfill

evan-cz:
## Problem Context

The existing backfill job was a one-time Kubernetes Job that ran immediately upon helm install but had no recurring schedule. Users needed both immediate execution on installation and configurable recurring runs for ongoing data collection.

## Implementation Approach

**Dual Resource Strategy**: Implemented both a CronJob for scheduled runs and an immediate Job for instant execution on helm install, using only Kubernetes- native features to ensure compatibility with all deployment methods.

**Single Template File**: Used a range loop with small if/else clauses to generate both resources from one template file, maximizing code reuse while handling the minimal differences between CronJob and Job specifications.

**Kubernetes-Native Features Only**: Avoided Helm-specific hooks and annotations to ensure compatibility with helm template, kubectl apply, ArgoCD, Flux, and other GitOps tools.

### Template Structure
- **Range Loop**: `{{- range $jobType := list "CronJob" "Job" }}` generates both resources
- **Context Handling**: Used `$` variable for proper template context in loops
- **Small Conditionals**: Minimal if/else clauses around only the differing parts
- **Container Spec**: Identical container configuration for both resources

### Resource Specifications
- **CronJob**: Uses `jobTemplate.spec.template` nesting with schedule, concurrency control
- **Job**: Uses direct `template` structure without automatic cleanup
- **CronJob**: Uses `jobTemplate.spec` with history limits for automatic cleanup
- **Naming**: CronJob gets simple name, Job gets unique ID for config change detection
- **Cleanup**: CronJob keeps only last 1 successful/failed job, immediate Job remains until manual deletion

### Configuration Options
- **Schedule**: Configurable via `initBackfillJob.schedule` (default: "0 */3 * * *")
- **Backward Compatibility**: All existing configuration options preserved
- **Schema Integration**: References official Kubernetes schema for schedule field

### Schema Tests
- **Valid Tests**: 4 test files covering various cron schedule formats
- **Invalid Tests**: 3 test files testing type validation (number, array, object)
- **Parallel Execution**: All tests pass with `make -j helm-test`

### Deployment Verification
- **CronJob Status**: Properly configured with 5-minute schedule for testing
- **Immediate Job**: Completed successfully in 4 seconds with proper cleanup
- **Template Rendering**: All templates render correctly with schedule override

### Code Quality
- **Go Tests**: All tests pass with parallel execution
- **Helm Tests**: All schema and template tests pass
- **Formatting**: All code properly formatted with no linting issues
- **Schema Validation**: New schedule field tests all pass

## Schema Enhancement

**Kubernetes Schema Integration**:
- Updated `initBackfillJob.schedule` to reference official Kubernetes schema
- Uses `$ref: "#/$defs/io.k8s.api.batch.v1.CronJobSpec/properties/schedule"`
- Provides authoritative description and documentation link
- Ensures consistency with Kubernetes API definitions

## Trade-offs and Decisions

**Dual Resource Approach**: Chose separate CronJob and Job over single CronJob with `startingDeadlineSeconds: 0` because the latter doesn't trigger immediate execution on creation, only on missed schedules. This approach evolved from an initial Helm hooks-based implementation that was replaced for GitOps compatibility.

**Kubernetes-Native Only**: Avoided Helm hooks to ensure compatibility with all deployment methods, using CronJob history limits for automatic cleanup.

**Single Template File**: Used range loop instead of separate files to maximize code reuse and maintainability while handling minimal differences efficiently.

**Schema Integration**: Referenced official Kubernetes schema instead of custom validation to ensure consistency and provide authoritative documentation.

## Future Considerations

- **Monitoring**: Consider adding metrics for CronJob execution success/failure
- **Alerting**: May want alerts for failed backfill jobs
- **Retention**: CronJob keeps only last 1 successful/failed job, immediate Job remains until deleted

## Files Changed

- `helm/templates/backfill-job.yaml`: Converted to dual-resource template with range loop
- `helm/templates/_helpers.tpl`: Added `initBackfillCronJobName` helper and updated naming
- `helm/values.schema.yaml`: Enhanced schedule field with Kubernetes schema reference
- `tests/helm/schema/`: Added 7 new schema test files for schedule validation

### DIFF
--- a/helm/templates/backfill-job.yaml
+++ b/helm/templates/backfill-job.yaml
@@ -1,76 +1,129 @@
 {{- if .Values.insightsController.enabled }}
 {{ $backFillValues := (include "cloudzero-agent.backFill" . | fromYaml) }}
 {{- if $backFillValues.enabled }}
+{{- /*
+  Backfill Job
+
+  The backfill job collects historical Kubernetes resource metadata for cost
+  allocation analysis. It runs immediately upon installation to capture existing
+  resources and then runs on a configurable schedule to maintain ongoing data
+  collection.
+
+  This template generates both a CronJob and an immediate Job using a range
+  loop:
+
+  The CronJob handles scheduled runs (default: every 12 hours). The Job runs
+  immediately upon helm install to capture existing resources without waiting
+  for the first scheduled run.
+
+  Helm hooks (post-install, hook-delete-policy) were tried, but rejected because
+  they require Helm-specific features that don't work with helm template,
+  kubectl apply, ArgoCD, Flux, or other GitOps tools
+
+  The implementation uses a range loop with small if/else clauses to maximize
+  code reuse while handling the minimal differences between CronJob and Job
+  specifications.
+*/ -}}
+{{- range $jobType := list "CronJob" "Job" }}
 apiVersion: batch/v1
-kind: Job
+kind: {{ $jobType }}
 metadata:
-  name: {{ include "cloudzero-agent.initBackfillJobName" . }}
-  namespace: {{ .Release.Namespace }}
-  {{- include "cloudzero-agent.generateAnnotations" (merge (deepCopy .Values.defaults.annotations) .Values.initBackfillJob.annotations) | nindent 2 }}
+  name: {{ if eq $jobType "CronJob" }}{{ include "cloudzero-agent.initBackfillCronJobName" $ }}{{ else }}{{ include "cloudzero-agent.initBackfillJobName" $ }}{{ end }}
+  namespace: {{ $.Release.Namespace }}
+  {{- include "cloudzero-agent.generateAnnotations" (merge (deepCopy $.Values.defaults.annotations) $.Values.initBackfillJob.annotations) | nindent 2 }}
   labels:
-    {{- include "cloudzero-agent.insightsController.labels" . | nindent 4 }}
+    {{- include "cloudzero-agent.insightsController.labels" $ | nindent 4 }}
+    job-type: backfill
+    {{- if eq $jobType "CronJob" }}
+    job-category: cronjob
+    {{- else }}
+    job-category: onetime
+    {{- end }}
 spec:
+  {{- if eq $jobType "CronJob" }}
+  schedule: {{ $.Values.components.webhookServer.backfill.schedule | default "0 */3 * * *" | quote }}
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 1
+  failedJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          name: {{ include "cloudzero-agent.initBackfillCronJobName" $ }}
+          namespace: {{ $.Release.Namespace }}
+          labels:
+            {{- include "cloudzero-agent.insightsController.initBackfillJob.matchLabels" $ | nindent 12 }}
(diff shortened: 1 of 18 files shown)


helm/values.yaml
```diff
@@ -368,6 +368,8 @@ components:
         cpu: "500m"
     # The backfill job component performs initial data backfilling.
     backfill:
+      # Schedule for the backfill CronJob. Defaults to every 3 hours. Use standard cron format.
+      schedule: "0 */3 * * *"
```
josephbarnett, line 372: Why 3 hours vs every 12? On large clusters this could be a lot...

JonParsons11350:
We're sure that this won't cause issues on the server side? I don't know enough to ask this question more intelligently, but, perhaps there's an assumptions that we only ever get a namespace label 1 time, and if we're suddenly seeing them come in multiple times a day, it will cause an issue..
I just want to make sure it's been thought about - no need for a detailed answer or explanation, just an acknowledgment is enough, thanks!

evan-cz:
> We're sure that this won't cause issues on the server side? I don't know enough to ask this question more intelligently, but, perhaps there's an assumptions that we only ever get a namespace label 1 time, and if we're suddenly seeing them come in multiple times a day, it will cause an issue.. I just want to make sure it's been thought about - no need for a detailed answer or explanation, just an acknowledgment is enough, thanks!

Nah. Good thought, but the server side should be okay. Though @josephbarnett can speak to that much better that I.

The load we're concerned with here is more on the client side; enumerating all the resources in the cluster isn't instantaneous, so this does put the cluster under a bit of extra load.

@josephbarnett I updated it to run every 12 hours by default.

helm/templates/backfill-job.yaml
```diff
@@ -1,76 +1,129 @@
...
 {{- if $backFillValues.enabled }}
+{{- /*
+  Backfill Job
+
+  The backfill job collects historical Kubernetes resource metadata for cost
+  allocation analysis. It runs immediately upon installation to capture existing
+  resources and then runs on a configurable schedule to maintain ongoing data
+  collection.
+
+  This template generates both a CronJob and an immediate Job using a range
```
dmepham, line 12: I may be misremembering, but don't cronjobs run immediately upon creation anyway? could we get best of both worlds by naming the cronjob with the values hash, and let it get recreated every config change?

helm/templates/backfill-job.yaml
```diff
@@ -1,76 +1,129 @@
...
+  loop:
+
+  The CronJob handles scheduled runs (default: every 3 hours). The Job runs
```
dmepham, line 15: I guess compared to usage data, this is probably such a small amount of data that 3 hours won't produce too much?

dmepham (approved):
I think I'm wrong about the cronjob running immediately upon creation thing, but worth asking I guess. lgtm!

helm/templates/backfill-job.yaml
evan-cz: Nope, I tried :(. I was hoping there would at least be a flag or something you could set, but I couldn't find anything.

helm/templates/backfill-job.yaml
evan-cz: nm, had to start it over anyways since CP-31335 merged, so I updated this too. Thanks!

PR #419, opened 2025-09-03
CP-32292: add Mermaid validation to linting pipeline

evan-cz:
## Why?

The project's Markdown documentation contains Mermaid diagrams for architecture documentation and troubleshooting guides, but there was no validation to catch syntax errors in these diagrams. This could lead to broken diagrams in documentation that would only be discovered when someone tried to render them.

## What

This implementation adds comprehensive Mermaid diagram validation to the existing linting pipeline using the official @mermaid-js/mermaid-cli tool. The solution uses Makefile pattern rules to efficiently process only files containing Mermaid diagrams, avoiding unnecessary work on files without diagrams.

## How Tested

`make lint`. Or `make lint-mermaid`.


docker/Dockerfile
```diff
@@ -60,11 +60,11 @@ RUN --mount=type=cache,target=/go/pkg/mod,id=gomod-$TARGETPLATFORM \
...
     REGENERATE=never \
-    TARGET_OS=$TARGETOS TARGET_ARCH=$TARGETARCH \
+    TARGET_OS="$TARGETOS" TARGET_ARCH="$TARGETARCH" \
     ENABLE_ZIG=true \
-    REVISION=${REVISION} \
-    TAG=${TAG} \
-    BUILD_TIME=${BUILD_TIME}
+    REVISION="${REVISION}" \
+    TAG="${TAG}" \
+    BUILD_TIME="${BUILD_TIME}"
```
evan-cz, line 67: Oh, this was included accidentally... I was playing with `hadolint` for linting Dockerfiles, and this was one of the things it took issue with. It also didn't like that we aren't pinning versions in the `apk` command, which seemed like overkill to me, and we don't have a good way to keep it up-to-date, so I decided not to add it for now.

This change, though, I think is a net win. Albeit a *very* small one.

dmepham (approved):

josephbarnett (approved):

PR #470, opened 2025-09-29
CP-33222: specialized cluster configuration testing infrastructure

evan-cz:
## Why?

We want to be able to test non-standard configurations in CI; for example, spin up a cluster in federated mode and verify that the KUTTL tests still pass.

## What

Previously this required creating an entirely new cluster in clusters/. This tweaks the helm invocation a bit so that, if `HELM_EXTRA_OVERRIDES` is provided when `helm-install`/`helm-install-current` is invoked, we use the file specified by it as an additional layer of overrides.

I also threw some extra stuff I've been working on in here, sorry... this translates primarily to more documentation, plus I added the markdownlint-cli2 linter to help keep our documentation tidy. Oh, and I also added a linter for github actions, which runs shell stuff through shellcheck, which uncovered a bunch of stuff (mainly around quoting).

## How Tested

I tested locally as best I could, the main question will be how CI feels about the changes...

### DIFF
--- a/.github/workflows/test-matrix-k8s.yaml
+++ b/.github/workflows/test-matrix-k8s.yaml
@@ -179,3 +179,131 @@ jobs:
           CLUSTER_NAME="kind-${{ matrix.kver.platform == 'linux/amd64' && 'amd64' || matrix.kver.platform == 'linux/arm64' && 'arm64' || matrix.kver.platform }}-${{ matrix.kver.k8s }}"
           # kind-test handles its own cleanup, but ensure cluster is deleted using Make target
           make kind-down CLUSTER_NAME=$CLUSTER_NAME || true
+
+  specialized-config-test:
+    name: config-test-${{ matrix.config.name }}
+    runs-on: ubuntu-latest
+    # These should match the permissions in the calling job.
+    permissions:
+      contents: read
+      packages: read
+    outputs:
+      config-test-status: ${{ job.status }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # Test specialized configurations that require different Helm values
+        config:
+          - name: "federated-mode"
+            description: "CloudZero Agent in federated mode with DaemonSet deployment"
+            overrides: |
+              defaults:
+                federation:
+                  enabled: true
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - uses: azure/setup-kubectl@v4
+
+      - name: Install Krew and Plugins
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 5
+          max_attempts: 4
+          command: |
+            (
+              set -x; cd "$(mktemp -d)" &&
+              OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+              ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
+              KREW="krew-${OS}_${ARCH}" &&
+              curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
+              tar zxvf "${KREW}.tar.gz" &&
+              ./"${KREW}" install krew
+            )
+            echo "$HOME/.krew/bin" >> "$GITHUB_PATH"
+            export PATH="$HOME/.krew/bin:$PATH"
+            kubectl krew install stern
+
+      - name: Install Dependencies
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 5
+          max_attempts: 4
+          command: |
+            make install-tools
+
+      - name: Create Specialized Configuration Override File
+        run: |
+          echo "Creating specialized configuration override file for ${{ matrix.config.name }}..."
+          # Create temporary file for extra overrides
+          EXTRA_OVERRIDES_FILE="${{ runner.temp }}/extra-overrides-${{ matrix.config.name }}.yaml"
+          cat <<'EOF' > "$EXTRA_OVERRIDES_FILE"
+          ${{ matrix.config.overrides }}
+          EOF
+          echo "HELM_EXTRA_OVERRIDES=$EXTRA_OVERRIDES_FILE" >> "$GITHUB_ENV"
+          echo "Created override file at: $EXTRA_OVERRIDES_FILE"
+          echo "Contents:"
+          cat "$EXTRA_OVERRIDES_FILE"
+
+      - name: Run Chart Tests with Specialized Configuration
+        env:
+          CLOUDZERO_DEV_API_KEY: ${{ secrets.CLOUDZERO_DEV_API_KEY }}
+        run: |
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -403,48 +413,6 @@ tests/kuttl/kubeconfig: ## Create kind cluster kubeconfig for testing
 
 # ----------- CI TESTING ------------
 
-# Generate the secrets file used by the `act` tool for local GitHub Action development.
```
jake-cloudzero, line 416: I assume this section moved somewhere else / into the runner manifests?

Makefile
```diff
@@ -569,6 +542,24 @@ define invoke-kubectl
 $(call get-kubeconfig-env) $(KUBECTL) $(call get-kubectx-arg,--context)
 endef
 
+# get-helm-extra-overrides - Generate additional --values argument if HELM_EXTRA_OVERRIDES is set
+#
+# This function enables specialized Helm chart configurations by allowing additional values files
```
jake-cloudzero, line 545: From what I understand, this is the main change correct?

Makefile
```diff
@@ -613,12 +605,17 @@ helm-install-current: ## Install chart with test image
...
 		--set components.agent.image.tag=dev-$(shell git rev-parse HEAD) \
 		$(NULL)
 
 .PHONY: helm-wait
 helm-wait: ## Wait for chart to be ready after installation
-	$(call invoke-kubectl) wait --for=condition=Available deployment/$(call get-cluster-property,.release)-cloudzero-agent-server --namespace $(call get-cluster-property,.namespace) --timeout=5m
+	$(call invoke-kubectl) wait --for=condition=Available \
+		--namespace $(call get-cluster-property,.namespace) \
+		--timeout=5m \
+		$(foreach deployment,cloudzero-agent-server cloudzero-agent-webhook-server aggregator cloudzero-state-metrics,deployment/$(call get-cluster-property,.release)-$(deployment)) \
```
jake-cloudzero, line 615: Cursed... Glad it works though!

.github/workflows/test-matrix-k8s.yaml
```diff
@@ -177,5 +176,133 @@ jobs:
...
+    outputs:
+      config-test-status: ${{ job.status }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # Test specialized configurations that require different Helm values
+        config:
+          - name: "federated-mode"
+            description: "CloudZero Agent in federated mode with DaemonSet deployment"
+            overrides: |
```
jake-cloudzero, line 199: So this is where the custom overrides are defined? Or are we referencing custom overrides files below?

.github/workflows/test-matrix-k8s.yaml
```diff
@@ -177,5 +176,133 @@ jobs:
...
+          command: |
+            make install-tools
+
+      - name: Create Specialized Configuration Override File
+        run: |
+          echo "Creating specialized configuration override file for ${{ matrix.config.name }}..."
+          # Create temporary file for extra overrides
+          EXTRA_OVERRIDES_FILE="${{ runner.temp }}/extra-overrides-${{ matrix.config.name }}.yaml"
+          cat <<'EOF' > "$EXTRA_OVERRIDES_FILE"
+          ${{ matrix.config.overrides }}
```
jake-cloudzero, line 246: On the above comment, I see... So here the overrides are injected into this step. Is there a reason why we inject in this way or would it just make more sense to define the overrides here?

jake-cloudzero: Or, does the action map over all instances of the config object? I see that it is a list, but I cannot find any code that references some sort of loop. Is this built into the action itself?

jake-cloudzero (commented):
Just so I am understanding properly, you basically added the ability to define different overrides files to test different specific configuration circumstances in our CI? Can you give a couple examples on how we do that in this PR? It seems that you add another configuration chunk in that CI file, but just want to be sure that is ALL we have to do.

Makefile
evan-cz: Sort of. Since we have config files for each cluster, you can dow just dump the API key into the cluster config file.

Makefile
evan-cz: Sort of. It's *part of* the main change... that one is actually *mostly* contained in a single commit, so it's probably easiest to just look at 94099466633b68ccebb7c6fe8bba9230b7e85eca to get an understanding of that change.

Makefile
evan-cz: 😆. Honestly, it's no less readable than the python version:

```python
[f"deployment/{get_cluster_property('.release')}-{deployment}" for deployment in ["cloudzero-agent-server", "cloudzero-agent-webhook-server", "aggregator", "cloudzero-state-metrics"]]
```

.github/workflows/test-matrix-k8s.yaml
evan-cz: Yep, these are the custom overrides which are referenced below.

.github/workflows/test-matrix-k8s.yaml
evan-cz: Yep, you're on the right track here! It's a build matrix. Here are some docs which do a way better job of explaining than I would: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations

Currently we only have one item in the list, but if you add another item it will create a job for that item, and there will be two jobs kicked off by this workflow. Add a third and you get three, and so on, without having to copy and paste the job again.

The previous job in this file is actually a great example of this; we have a single `job` in the YAML, but then a matrix which builds a bunch of different k8s versions. And we end up with like 11 different jobs.

FWIW it's well worth reading that doc to understand what it does... it's *really* powerful when you're building more advanced CI/CD pipelines.

evan-cz:
> Just so I am understanding properly, you basically added the ability to define different overrides files to test different specific configuration circumstances in our CI? Can you give a couple examples on how we do that in this PR? It seems that you add another configuration chunk in that CI file, but just want to be sure that is ALL we have to do.

Yep, that's pretty much it. If you want to add another configuration to CI, you would just add another item to [this list](https://github.com/Cloudzero/cloudzero-agent/blob/311cdb2c40cd0d0abd43198003a939086041eb03/.github/workflows/test-matrix-k8s.yaml#L196-L201). If you want to do it locally, just drop the additional overrides file wherever you want, then do something like `make helm-install HELM_EXTRA_OVERRIDES=path/to/overrides/file.yaml CLUSTER_NAME=mycluster` (assuming you have a `clusters/mycluster.yaml` set up, which you definitely should do now!).

In the future, we *could* integrate this with other parts of the tests; for example, replacing / supplementing the overrides files in tests/helm/template/. But that presents some obstacles I'd want to think through first.

We could/should also add additional configurations; one for using cert-manager comes to mind, but really anything would work. And the exact same KUTTL tests will run against them as run against all the different k8s versions, and the replicated clusters.

Makefile
jake-cloudzero: I always found Makefile semantics crazy 😆 But yes I suppose so...

.github/workflows/test-matrix-k8s.yaml
jake-cloudzero: Oh wow! Deep into the world if Github CI... Very cool thanks for the doc.

jake-cloudzero (approved):

PR #476, opened 2025-10-02
Support post-processing of helm templating

megrehn-sf:
### Description

This regex validation prevents the use of this chart with post helm-templating secrets engines. 
Example: The argocd plugin for Vault when used with helm charts. https://argocd-vault-plugin.readthedocs.io/en/stable/

### References

The vault plugin for argocd applies secrets templating post helm templating. 
No obvious reason to include a regex validation here.

<code>{"dir":"/tmp/_cmp_server/c26aa731-ed23-4bd9-a549-3bc43995ac04/charts/cloudzero-agent","execID":"70d36","level":"info","msg":"sh -c \"RELEASE_NAME=\\\"${ARGOCD_ENV_HELM_RELEASE_NAME:-$ARGOCD_APP_NAME}\\\";\\necho \\\"$ARGOCD_ENV_HELM_VALUES\\\" \u003e ARGOCD_ENV_HELM_VALUES.yaml |\\nhelm template $RELEASE_NAME -n $ARGOCD_APP_NAMESPACE --values ARGOCD_ENV_HELM_VALUES.yaml . |\\nargocd-vault-plugin generate -\\n\"","time":"2025-10-01T23:39:34Z"}
{"command":{"command":["sh","-c","RELEASE_NAME=\"${ARGOCD_ENV_HELM_RELEASE_NAME:-$ARGOCD_APP_NAME}\";\necho \"$ARGOCD_ENV_HELM_VALUES\" \u003e ARGOCD_ENV_HELM_VALUES.yaml |\nhelm template $RELEASE_NAME -n $ARGOCD_APP_NAMESPACE --values ARGOCD_ENV_HELM_VALUES.yaml . |\nargocd-vault-plugin generate -\n"]},"execID":"70d36","level":"warning","msg":"Plugin command returned zero output","stderr":"Error: execution error at (cloudzero-agent/templates/aggregator-secret.yaml:13:14): The provided apiKey is invalid. Check that the provided value from apiKey matches exactly what is found in the CloudZero admin page.\n\nUse --debug flag to render out invalid YAML\n","time":"2025-10-01T23:39:35Z"}</code>

### Testing

No effectual change to the chart installation. Test with install

<code>helm upgrade --install cloudzero \
    --repo https://cloudzero.github.io/cloudzero-charts cloudzero-agent \
    --namespace cloudzero --create-namespace \
    --set apiKey=<CLOUDZERO_API_KEY>
    --set clusterName=<NAME_OF_YOUR_CLUSTER>
</code>

PR #477, opened 2025-10-02
CP-33383: move apiKey/existingSecretName validation to JSON Schema

evan-cz:
## Why?

We have received a bug report (#476) about the regular expression-based validation in the template for the API key. Apparently, this method of validation causes problems for certain tools.

Huge thanks to @megrehn-sf for doing the the initial solution and doing the legwork on this; I'm just building on that patch here. 

## What

Move the validation from the template into the JSON schema. Furthermore, replace the other API key and existing secret name validation to ensure that only one or the other is set. Finally, add tests for everything.

## How Tested

Additional helm unittest tests and schema tests.

### DIFF
--- a/helm/templates/aggregator-secret.yaml
+++ b/helm/templates/aggregator-secret.yaml
@@ -9,8 +9,5 @@ metadata:
   namespace: {{ include "cloudzero-agent.namespace" . }}
 data:
   {{ .Values.serverConfig.containerSecretFileName }}: {{- $apiKey := .Values.apiKey | toString }}
-          {{- if not (regexMatch "^[a-zA-Z0-9-_.~!*'();]+$" $apiKey) }}
-          {{- fail "The provided apiKey is invalid. Check that the provided value from apiKey matches exactly what is found in the CloudZero admin page." }}
-          {{- end }}
           {{ $apiKey | b64enc | quote }}
 {{- end }}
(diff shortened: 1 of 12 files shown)


roberthocking (approved):

megrehn-sf:
Hi @evan-cz , while I can agree this is a more helm-like way to do values validation. I don't believe it will address the issue in https://github.com/Cloudzero/cloudzero-agent/pull/476 as these values are still validated during helm template operations, but the reported bug is on the vault secrets plugin for argocd which does post processing on the generated manifests.

Edit: I can give it a try though, it does look like there's a `--skip-schema-validation` flag i can add for this applicationset which may be able to workaround for my case.

evan-cz:
I don't have any experience the vault plugin for ArgoCD, but perhaps it would be better to use the `existingSecretName` field, with the value set to the name of the secret that plugin is creating for you? That way there is no need to patch the manifest after helm processes it.

Basically, if you set `apiKey` then the chart will create a K8s secret for you with that content. However, if you specify a resource name with `existingSecretName` the chart will just assume that the secret already exists and use that.

`--skip-schema-validation` could be an option; I know ArgoCD has support for this in the UI. But I'd recommend avoiding it if possible; the JSON Schema validation honestly does catch a lot of issues, it's a great check to have in place.

PR #485, opened 2025-10-06
CP-33514: add server svc port protocol

jake-cloudzero:
A PR to include #465 from a private contributor until we can fix CI to work for external contributors.


evan-cz:
Do we have an associated jira issue yet? We need one... (SOC 2 isn't really designed with external contributors in mind)

evan-cz (approved):

PR #512, opened 2025-10-27
CP-27876: Update Helm NOTES.txt with post-install guidance

amfelso:
# Why?

Implementing [a suggestion from Thomas](https://cloudzero.atlassian.net/browse/CP-27876) to properly set expectations post-install

# What

Added a link to cluster list page and some verbiage on timeline

# How Tested

Local


dmepham (approved):
nice!

helm/templates/NOTES.txt
```diff
@@ -47,4 +47,10 @@ Please refer to the documentation for guidance on `kubeStateMetrics` settings.
 
                       Install Successful
 
-Your data will be available in the CloudZero portal after 48 hours. 
+→ Next Steps:
+
+1. Your cluster should appear within the next hour at:
+   https://app.cloudzero.com/organization/k8s-integration.
```
evan-cz: One *very* small nit; I think we should either get rid of the period or enclose the URL in angle brackets (I'd probably go with the former in this case, but I don't have a strong opinion)... I'm not sure how different tools will parse that, but technically I believe URLs can end with a dot, and I wouldn't be shocked if some terminals included it when highlighting the link.

amfelso: No problem, will make the change!

PR #548, opened 2025-11-19
CP-35266: remove some unused files to free disk space in CI

evan-cz:
# Why?

The disk being almost full was causing some tests to fail in CI.

# What

This just deletes some stuff we don't use (like .NET and GHC) before running the smoke tests.

# How Tested

Push, wait (impatiently) for CI to run.

### DIFF
--- a/.github/workflows/golang-ci.yml
+++ b/.github/workflows/golang-ci.yml
@@ -212,6 +212,28 @@ jobs:
       - name: Install dependencies in tests directory
         run: go -C tests mod tidy
 
+      # Free up disk space to prevent Docker overlay FS from exceeding 85%
+      - name: Free up disk space
+        run: |
+          echo "=== Cleaning up disk space ==="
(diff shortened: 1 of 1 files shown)


.github/workflows/golang-ci.yml
```diff
@@ -212,6 +212,28 @@ jobs:
...
+          # Remove Docker images and build cache
+          docker system prune -af --volumes
+
+          # Clean apt cache
+          sudo apt-get clean
+          sudo rm -rf /var/lib/apt/lists/*
+
+          # Remove large tools we don't need
+          # Android SDK, .NET, etc. are often pre-installed but unused
+          sudo rm -rf /usr/share/dotnet
```
roberthocking, line 229: rm -rf dotnet - wiser words have never been said. kill it with fire!

roberthocking (approved):

PR #582, opened 2025-12-04
CP-35660: add script to list all labels/annotations in a cluster

evan-cz:
# Why?

Sometimes the people installing the chart don't really know what labels / annotations are in use in the chart, or how commonly used they are. This makes it difficult to know which labels to collect.

# What

This provides a convenient way to get a list of all labels and/or annotations in use in a cluster, including how frequently they occur. Optionally, it can also include all of the values of those labels and annotations, including their frequency.

# How Tested

* `./scripts/kube-list-labels-annotations.sh` -- Show labels
* `./scripts/kube-list-labels-annotations.sh -v` -- Show labels and values
* `./scripts/kube-list-labels-annotations.sh -a` -- Show annotations
* `./scripts/kube-list-labels-annotations.sh -a -v` -- Show annotations and values
* `./scripts/kube-list-labels-annotations.sh -a -l` -- Show labels and annotations
* `./scripts/kube-list-labels-annotations.sh -a -l -v` -- Show labels, annotations, and the values of each
* `./scripts/kube-list-labels-annotations.sh -- --context my-kubectl-ctx` -- Everything after the ` -- ` gets dropped into the `kubectl` command

### DIFF
--- a/scripts/kube-list-labels-annotations.sh
+++ b/scripts/kube-list-labels-annotations.sh
@@ -1,0 +1,202 @@
+#!/usr/bin/env sh
+#
+# kube-list-labels-annotations.sh
+#
+# List all labels and/or annotations from all resources in a Kubernetes cluster,
+# sorted by frequency with counts.
+#
+# This script queries ALL resource types in the cluster (pods, services,
+# deployments, configmaps, secrets, nodes, etc.) and aggregates their labels
+# and/or annotations to show which ones are most commonly used.
+#
+# USAGE
+#   ./scripts/kube-list-labels-annotations.sh [options] [-- kubectl-args...]
+#
+# OPTIONS
+#   -l, --labels       Show labels (default if neither -l nor -a specified)
+#   -a, --annotations  Show annotations
+#   -v, --values       Show values for each key (hierarchical output)
+#   -h, --help         Show this help message
+#
+#   Any arguments after '--' are passed directly to kubectl. This is useful
+#   for specifying context, namespace restrictions, or authentication options.
+#
+# OUTPUT FORMAT
+#   Without -v (keys only):
+#      COUNT LABEL_KEY
+#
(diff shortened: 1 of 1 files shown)


scripts/kube-list-labels-annotations.sh
```diff
@@ -0,0 +1,185 @@
+#!/bin/bash
```
josephbarnett: I generally try to use `env` which is super common and will use the shell or fallback. 

```suggestion
#!/usr/bin/env bash
```

scripts/kube-list-labels-annotations.sh
```diff
@@ -0,0 +1,185 @@
...
+#
+# kube-list-labels-annotations.sh
+#
+# List all labels and/or annotations from all resources in a Kubernetes cluster,
+# sorted by frequency with counts.
+#
+# This script queries ALL resource types in the cluster (pods, services,
+# deployments, configmaps, secrets, nodes, etc.) and aggregates their labels
+# and/or annotations to show which ones are most commonly used.
+#
```
josephbarnett, line 11: Add a small DEPS section here listing:

1. kubectl
2. awk (sometimes this is not always available).

scripts/kube-list-labels-annotations.sh
josephbarnett: After reviewing this, I don't see anything that prevents you from using `sh` instead of `bash` (ie. associative arrays) I still recommend using `env` to for shell selection.

josephbarnett (approved):
NICE SCRIPT! A couple of nits...

scripts/kube-list-labels-annotations.sh
evan-cz: Switched to `/bin/sh` since I removed the bash dependency... I could do `/usr/bin/env sh` I guess, but that feels unnecessary...

PR #605, opened 2025-12-18
CP-32193: Add Istio integration with runtime validation

evan-cz:
Istio service mesh deployments create two challenges for the CloudZero Agent: (1) strict mTLS can conflict with the webhook's self-signed TLS certificates, and (2) cross-cluster load balancing can route metrics to the wrong aggregator, corrupting cost attribution data. This change adds automatic Istio detection, runtime validation, and comprehensive documentation.

Functional Change:

Before: Istio support required manual configuration via suppressIstioAnnotations. No runtime validation existed for cross-cluster load balancing risks. Documentation was incomplete and didn't explain the underlying problems.

After: The chart auto-detects Istio via CRD presence, applies port exclusion annotations automatically, and the validator detects both sidecar and ambient modes at runtime, warning when cluster ID configuration is incorrect or missing.

Solution:

1. Added `integrations.istio.enabled` (null/true/false) for auto-detect or explicit control, replacing the old `suppressIstioAnnotations` approach

2. Added `integrations.istio.clusterID` for traffic fencing in multicluster meshes, with fallback to `clusterName` for automatic configuration in sidecar mode

3. Created new Istio diagnostic provider (`app/domain/diagnostic/istio/`) that:
   - Detects sidecar mode by querying localhost:15000 (Envoy admin API)
   - Detects ambient mode via Downward API (ambient.istio.io/redirection annotation)
   - In sidecar mode, validates effective cluster ID (clusterID || clusterName) matches what Istio reports, ensuring the DestinationRule will work correctly
   - Requires explicit clusterID in ambient mode (cannot validate without sidecar)

4. Added Helm helpers for Istio detection and cluster ID resolution with detailed documentation explaining the port number differences (443 vs 8443)

5. Completely rewrote helm/docs/istio.md to explain the two problems (mTLS conflict and cross-cluster LB) with Mermaid diagrams, solutions, and configuration guide

Validation:

- Added comprehensive unit tests for Istio diagnostic provider (istio_test.go) covering sidecar mode, ambient mode, cluster ID validation, and edge cases
- Added Helm unittest suite (istio_integration_test.yaml) with 33 tests covering DestinationRule/VirtualService generation, port exclusion annotations, API version selection, and cluster ID configuration
- Added schema validation tests for integrations.istio.enabled and clusterID
- Added Istio template test to verify full manifest generation with Istio enabled
- Manual testing pending on Istio-enabled clusters (sidecar and ambient modes)


app/domain/diagnostic/istio/istio.go
```diff
@@ -0,0 +1,444 @@
+// SPDX-FileCopyrightText: Copyright (c) 2016-2025, CloudZero, Inc. or its affiliates. All Rights Reserved.
```
dmepham, line 1: I don't know if this is intentional or if it is an issue, but I think all the other diagnostics are named in the style `<descriptive-name>/check.go`, whereas this is `<descriptive-name>/descriptive-name.go`. just noting in case

app/domain/diagnostic/istio/istio.go
```diff
@@ -0,0 +1,444 @@
...
+				regionPart = strings.TrimSpace(regionPart)
+				if regionPart != "" {
+					c.logger.Debugf("Found region '%s' for service %s", regionPart, servicePattern)
+					regions[regionPart] = struct{}{}
+				}
+			}
+		}
+	}
+
+	// If we have more than one distinct region, cross-cluster LB is active
```
dmepham, line 424: would this miss cross-clusters in the same region?

app/functions/helmless/default-values.yaml
```diff
@@ -206,6 +206,42 @@ defaults:
...
+    # - false: Force Istio integration disabled
+    enabled: null
+    # Istio cluster ID for multicluster mesh environments.
+    #
+    # In an Istio multicluster mesh, each cluster has a unique cluster ID that
+    # identifies it within the mesh. This value is used by the DestinationRule
+    # to pin traffic to pods in the local cluster using the
+    # `topology.istio.io/cluster` label that Istio automatically adds to pods.
+    #
+    # If not set, falls back to the top-level `clusterName` value. At least one
```
dmepham, line 227: I wonder if it would make sense to have the fallback be a clusterName-cloudAccountId-region combination?

helm/docs/istio.md
```diff
@@ -1,80 +1,224 @@
...
-insightsController:
-  server:
-    suppressIstioAnnotations: true
+- **Sidecar mode** injects an Envoy proxy container into each pod. All traffic passes through this sidecar, which handles mTLS termination and origination.
+
+- **Ambient mode** (Istio 1.18+) uses a node-level proxy called ztunnel instead of per-pod sidecars. Traffic is transparently redirected through ztunnel without modifying pod definitions.
+
+### How the Webhook Server Works
+
+The CloudZero webhook server receives notifications from the Kubernetes API server when pods are created or updated. It extracts metadata (labels, annotations, ownership) for cost attribution purposes and sends this data to the aggregator.
```
dmepham, line 19: nit: maybe `when pods are created or updated` -> `when resources are created or updated`? maybe it would be too verbose to list all the things that we get updates for

dmepham (approved):
some minor questions, but lgtm!

app/functions/helmless/default-values.yaml
evan-cz: The issue with this is that the fallback actually has to match the Istio cluster ID as configured by the customer; we don't get to choose the format. I figure the most common thing here is going to be for the customers to configure the Istio cluster ID to have the same value as they configured the cluster name in their cloud provider, but we can't be certain.

When Istio is running in Sidecar mode, we can query the Sidecar to validate that the name is set correctly because we can access the Sidecar's API on `localhost:15000`, and simply fail if the values don't match. Unfortunately, though, since we need the cluster ID when Helm is generating the template (and don't have access to the sidecar until it's already deployed), we can't just automatically detect the correct value and use it. So just detecting an incorrect value is about the best we can do. 

In ambient mode we don't have a sidecar to query, so there's no way to detect/validate the cluster ID. Which is why I've made it a requirement that if Istio is in ambient mode, the customer must explicitly provide a cluster ID instead of falling back on the cluster name and crossing our fingers.

app/domain/diagnostic/istio/istio.go
evan-cz: Good catch. IMHO that's actually something of an anti-pattern, but we should be consistent, so I'll update it.

app/domain/diagnostic/istio/istio.go
evan-cz: Good question! I don't *think* so; it's just appending to an array not setting values in a map, but I'll spin up a new cluster to check...

app/functions/helmless/default-values.yaml
dmepham: gotcha, makes sense, thanks!

app/domain/diagnostic/istio/istio.go
evan-cz: Good catch. You're right that the region-based approach would miss same-region clusters.

**Important context**: This detection is purely diagnostic/informational - the actual traffic fencing protection comes from the DestinationRule that pins traffic using the `topology.istio.io/cluster` label. Even if detection failed, traffic would still be correctly fenced. That said, misleading log messages aren't great.

Upon further examination, I found a more reliable detection method: comparing the `local-cluster` subset endpoints against all endpoints in the Envoy output:

```
# Local-cluster subset (our DestinationRule):
outbound|80|local-cluster|aggregator.cza.svc.cluster.local::10.84.0.10:8080

# All endpoints (includes remote clusters):
outbound|80||aggregator.cza.svc.cluster.local::10.84.0.10:8080
outbound|80||aggregator.cza.svc.cluster.local::10.36.6.3:8080  <- remote
```

If there are IPs in the all-endpoints list that aren't in the local-cluster subset, those are remote endpoints from other clusters - regardless of their region. This approach works because it uses the same mechanism as the DestinationRule itself.

I've updated the code to use this subset-based detection with comprehensive tests, including `TestParseEnvoyResponse_SameRegion_DifferentClusters` that specifically validates this scenario.

This is now tested with a second GKE cluster in the same region, and everything seems to work as expected.

PR #606, opened 2025-12-18
CP-35716: fix defaults.image.pullSecrets not applying to workload templates

evan-cz:
While adding comprehensive per-type unit tests for defaults.* properties, discovered that defaults.image.pullSecrets was not being applied to most workload templates. Only config-loader-job.yaml and helmless-job.yaml were using the correct generateImagePullSecrets helper; six other templates used legacy helpers that did not fall back to defaults.image.pullSecrets.

Functional Change:

Before: Setting defaults.image.pullSecrets had no effect on agent-deploy, aggregator-deploy, agent-daemonset, webhook-deploy, backfill-job, or init-cert-job templates. Users had to set the deprecated top-level imagePullSecrets or configure each component individually.

After: Setting defaults.image.pullSecrets applies to all workload templates. Backwards compatibility with the deprecated imagePullSecrets is preserved.

Root Cause:

Six templates were using legacy imagePullSecrets helpers that had different fallback chains:

1. cloudzero-agent.server.imagePullSecrets - only checked .Values.imagePullSecrets
2. cloudzero-agent.insightsController.server.imagePullSecrets - checked insightsController.server.imagePullSecrets -> imagePullSecrets
3. cloudzero-agent.initBackfillJob.imagePullSecrets - checked backFillValues.imagePullSecrets -> insightsController.server.imagePullSecrets -> imagePullSecrets
4. cloudzero-agent.initCertJob.imagePullSecrets - similar chain

None of these helpers included defaults.image.pullSecrets in their fallback chain.

Solution:

1. Updated generateImagePullSecrets helper (_helpers.tpl:1137-1142) to include deprecated imagePullSecrets as final fallback: .image.pullSecrets | default .root.Values.defaults.image.pullSecrets | default .root.Values.imagePullSecrets

2. Updated six templates to use generateImagePullSecrets:
   - agent-deploy.yaml (line 310)
   - aggregator-deploy.yaml (line 159)
   - agent-daemonset.yaml (line 190)
   - webhook-deploy.yaml (line 86)
   - backfill-job.yaml (line 122)
   - init-cert-job.yaml (line 54)

3. Also fixed aggregator-service.yaml to use generateLabels for consistency with other templates (removes app.kubernetes.io/component which was inconsistently applied across resources).

Validation:

- All 395 Helm unit tests pass (8 new tests added for defaults.image.pullSecrets)
- New tests verify defaults.image.pullSecrets applies to Deployment, DaemonSet, Job, and CronJob resources
- New tests verify backwards compatibility with deprecated imagePullSecrets
- Manual helm template verification confirms imagePullSecrets renders correctly when defaults.image.pullSecrets is set


tests/helm/template/manifest.yaml
```diff
@@ -2071,10 +2071,9 @@ metadata:
   namespace: cz-agent
   name: cz-agent-cz-aggregator
   labels:
-    app.kubernetes.io/component: aggregator
     app.kubernetes.io/instance: cz-agent
     app.kubernetes.io/managed-by: Helm
-    app.kubernetes.io/name: cz-agent-cz-aggregator
+    app.kubernetes.io/name: aggregator
```
dmepham, line 2076: maybe this is just a change we have in this chart that I didn't know about, but I think typically we'd expect
```yaml
app.kubernetes.io/name: cloudzero-agent
app.kubernetes.io/component: aggregator
```
based on the common [helm label suggestions](https://helm.sh/docs/chart_best_practices/labels/).

but, I also see that we're using `app.kubernetes.io/name: aggregator` in the spec.selector below. I think this changed [here](https://github.com/Cloudzero/cloudzero-agent/commit/c7a0b6fffb341d848075588fcc29331c3ab147c4). would it be worth considering reverting it? I think it's probably not a big deal, but is a convention. I think some of our troubleshooting guides might reference `component`, but other than that I don't think it would cause issues

evan-cz: Yeah, that was an intentional change. See https://github.com/Cloudzero/cloudzero-agent/blob/145c23988a761dca648b1def67dc39bfca3e1a07/helm/templates/_helpers.tpl#L894-L897

The documentation kind of sucks, but I did a lot of research on what the different labels are actually supposed to be, and updated the chart to use the correct values. So I think the proper change here would actually be to update the documentation to use name instead of component. I'll make sure to do a pass on that soon.

dmepham: got it, I missed that, thanks

dmepham (approved):

PR #648, opened 2026-02-03
CP-37371: Use distroless Prometheus image by default

evan-cz:
Prometheus 3.10.0 introduced an official distroless image variant that eliminates shells, package managers, and other utilities from the container. This significantly reduces the attack surface of the Prometheus container, addressing security concerns around the nodes/proxy RBAC permission (see docs/wiki/The-nodes-proxy-Problem.md).

Functional Change:

Before: The Helm chart defaulted to the standard Prometheus image (busybox-based, includes shell and utilities) at version v3.7.3. The Prometheus image tag was resolved by the `cloudzero-agent.prometheusImageTag` helper with an inline fallback chain duplicated across templates.

After: The Helm chart defaults to the distroless Prometheus image at version v3.10.0. The image tag is resolved by a new `cloudzero-agent.Values.components.prometheus.image.tag` helper that appends "-distroless" to Chart.AppVersion when no explicit tag is set. The deprecated server.image.tag compat override is handled by generateImage's compat layer, not the helper itself.

Solution:

1. Bumped Chart.AppVersion from v3.7.3 to v3.10.0 in helm/Chart.yaml

2. Added `cloudzero-agent.Values.components.prometheus.image.tag` helper in helm/templates/_helpers.tpl that resolves components.prometheus.image.tag with a fallback to `{Chart.AppVersion}-distroless`

3. Removed the old `cloudzero-agent.prometheusImageTag` helper and updated all call sites (agent-deploy.yaml, agent-daemonset.yaml, prometheusAgentFlag) to use the new helper

4. Removed outdated inline comments about version-specific distroless logic from agent-deploy.yaml and agent-daemonset.yaml

5. Regenerated Helm template test snapshots (alloy.yaml, cert-manager.yaml, federated.yaml, istio.yaml, manifest.yaml) to reflect the new default image tag

Validation:

- Helm template snapshots regenerated and verified

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -8,7 +8,7 @@ kubeVersion: ">= 1.21.0-0"
 maintainers:
   - name: CloudZero
     email: support@cloudzero.com
-appVersion: "v3.7.3"
+appVersion: "v3.10.0"
 dependencies:
   - name: kube-state-metrics
     version: "5.36.*"
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -1390,6 +1390,23 @@ prometheus.yml
 {{- end -}}
 
 {{/*
+Resolve the Prometheus image tag.
+
+Resolves the tag from components.prometheus.image.tag, falling back to
+Chart.AppVersion with "-distroless" appended to use the official distroless
+image variant (no shell, minimal attack surface).
+
+Note: the deprecated server.image.tag compat override is handled by
+generateImage's compat layer at the call site, not here.
+
+Usage: {{ include "cloudzero-agent.Values.components.prometheus.image.tag" . }}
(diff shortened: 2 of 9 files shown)


.github/workflows/golang-ci.yml
```diff
@@ -141,6 +141,10 @@ jobs:
         run: git diff --exit-code --color
 
   validate-dependabot: # Validate Dependabot configuration
+    # Disabled: marocchino/validate-dependabot@v3 fails due to a schemastore.org
+    # schema change (base.json#/definitions/timezone reference can't be resolved).
+    # See https://json.schemastore.org/dependabot-2.0.json
+    if: github.event_name == 'disabled'
```
evan-cz, line 147: Not really part of this, but the action is failing and this uncorks CI

roberthocking (approved):

PR #785, opened 2026-04-22
CP-40647: add KUTTL regression test for Alloy clustered mode

evan-cz:
CP-40307 fixed CrashLoopBackOff on clustered-mode installs caused by the distroless agent image having no writable /tmp for Alloy's --storage.path=/tmp/alloy. That regression was reported by customers post-release rather than caught in testing. The fix commit added helm-unittest coverage, which validates template output, but never actually runs the container — meaning a similar runtime-only regression could still slip through.

Functional Change:

Before: Clustered mode had no end-to-end test. A regression that only manifests when the container starts (permission denied, missing mount, misconfigured arg) could pass all pre-merge checks

After: A dedicated KUTTL suite installs the chart in clustered mode, waits for the Alloy pod to become Ready, and asserts the CP-40307-specific log signature is absent.

Implementation Approach:

The suite is self-contained and installs its own helm release in a dedicated namespace rather than reconfiguring the default cz-agent install that other KUTTL suites run against. This avoids the side effects of switching the chart between agent and clustered modes mid-cycle (Server Deployment → Alloy StatefulSet reshape) and keeps the new suite additive.

1. tests/kuttl/alloy-clustered-test/kuttl-test.yaml — TestSuite, deleteNamespace: false so the step 99 helm uninstall can run first.
2. steps/01-install-clustered-chart.yaml — helm upgrade --install with components.agent.mode=clustered and --wait --timeout=3m. The dev image tag mirrors the helm-install-current Makefile target.
3. steps/02-verify-alloy-ready.yaml — kubectl wait for pod Ready, assert Alloy container restartCount == 0, grep logs for the "mkdir /tmp: permission denied" signature and fail if present.
4. steps/99-uninstall.yaml — best-effort helm uninstall and namespace deletion so subsequent suites and kind-test teardown are unaffected.

No Makefile or CI changes are needed. helm-test-kuttl auto-discovers suites via $(wildcard tests/kuttl/*/kuttl-test.yaml), and docker-build already invokes test-matrix-k8s.yaml (which runs make kind-test) across the k8s version matrix on every PR and in the merge queue.

Validation:

- YAML syntax verified for all four step files.
- make -n tests/kuttl/alloy-clustered-test/run confirms the suite is discovered by the existing Makefile machinery.
- Full make kind-test run and revert-the-fix negative test to be executed in CI (this branch's purpose) and locally prior to merge.

### DIFF
--- a/tests/kuttl/alloy-clustered-test/kuttl-test.yaml
+++ b/tests/kuttl/alloy-clustered-test/kuttl-test.yaml
@@ -1,0 +1,19 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+metadata:
+  name: alloy-clustered-test
+  annotations:
+    kuttl.dev/type: "test"
+spec:
+  testDirs:
+    - ./steps
+  timeout: 300
+  parallel: 1
+  reportFormat: JSON
+  reportName: alloy-clustered-test-report
+  # The suite installs its own helm release in a dedicated namespace and
+  # uninstalls it in step 99, so we do not let kuttl delete the namespace
+  # (that would leave helm release state orphaned).
+  deleteNamespace: false
+  namespaceLabels:
+    test: alloy-clustered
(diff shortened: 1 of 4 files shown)


josephbarnett (approved):
Approved, but I know nothing of these configurations and can't verify they are correct

evan-cz:
> Approved, but I know nothing of these configurations and can't verify they are correct

No worries, this is just about adding additional testing to verify the that the chart at least installs and starts in clustered mode.

PR #802, opened 2026-04-29
CP-40824: Add automemlimit to prevent GC-induced OOM kills

evan-cz:
A customer reported that their backfill jobs are being OOM-killed. The
backfill job enumerates Kubernetes resources and streams them to the
collector — it shouldn't retain much state, yet it was exceeding its
512 MiB memory limit.

Investigation confirmed the root cause is Go's GC being too lazy: without
GOMEMLIMIT, the runtime only triggers collection when the heap doubles,
allowing RSS to drift past the cgroup limit before the kernel kills the
pod. Short-lived allocation churn (K8s API responses, protobuf
serialization, HTTP request bodies) accumulates faster than the GC
reclaims.

Functional Change:

Before: All agent binaries ran without GOMEMLIMIT. Go's GC had no awareness
of the pod's cgroup memory budget and would allow RSS to grow unchecked
until the kernel OOM-killed the process.

After: All binaries import automemlimit, which reads the cgroup v1/v2
memory limit at startup and sets GOMEMLIMIT to 90% of that value. The GC
now collects aggressively before RSS reaches the cgroup limit.

Root Cause:

Go's default GC heuristic targets 2x the live heap before triggering
collection. For the backfill job the live working set is modest, but the GC
runs only around 160 times across the entire run regardless of workload
size. Without GOMEMLIMIT, RSS drifts upward with allocation churn and
exceeds the cgroup limit before the GC bothers to collect.

With GOMEMLIMIT set (e.g., to 461 MiB = 90% of 512 MiB), the GC runs
whenever the heap approaches the limit, keeping RSS well below the cgroup
threshold.

Solution:

1. Added github.com/KimMachineGun/automemlimit v0.7.5 as a dependency.
   At process startup (via init()), this library reads the cgroup memory
   limit and calls debug.SetMemoryLimit() at 90% of that value.

2. Added a blank import of automemlimit to all eight binary entry points:
   webhook, collector, shipper, agent-inspector, agent-validator,
   certifik8s, cluster-config, regurgitator.

3. Added -backfill-no-wait flag to the webhook binary to skip waiting for
   dependent services, which the stress test harness needs.

4. Added out-of-process stress test harness (tests/stress/backfill/) that
   spawns the webhook binary against a fake K8s API server and collector
   sink, measuring RSS externally.

automemlimit behavior:
- On Linux with a cgroup limit: sets GOMEMLIMIT to 90% of the limit.
- On Linux without a cgroup limit: logs "memory is not limited" and skips.
- On non-Linux (macOS dev machines): logs an error and skips.
- If GOMEMLIMIT is already set via env var: skips (operator override).
- If AUTOMEMLIMIT=off is set via env var: disables entirely.

Validation:

Out-of-process RSS scaling (no GOMEMLIMIT, namespaces + pods only):

```
  Resources  Peak RSS   Sink Reqs  Duration
  11k         60 MiB         196   1m39s
  22k         66 MiB         397   3m23s
  105k       116 MiB         997   8m23s
  255k       193 MiB         997   8m23s
  510k       336 MiB       1,997   16m43s
  1.01M      597 MiB       1,997   16m44s
```

Memory grows linearly at approximately 0.5 KiB/resource. The default
512 MiB limit is sufficient for clusters up to about 800k resources. The
OOM kills were caused by GC laziness, not by retained state.

All existing unit tests pass (make test). All eight binaries build
(make build).

### DIFF
--- a/app/functions/agent-inspector/main.go
+++ b/app/functions/agent-inspector/main.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"time"
 
+	_ "github.com/KimMachineGun/automemlimit"
 	"github.com/cloudzero/cloudzero-agent/app/inspector"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
--- a/app/functions/agent-validator/main.go
+++ b/app/functions/agent-validator/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 
+	_ "github.com/KimMachineGun/automemlimit"
 	"github.com/cloudzero/cloudzero-agent/app/build"
 	configcmd "github.com/cloudzero/cloudzero-agent/app/functions/agent-validator/config"
 	diagcmd "github.com/cloudzero/cloudzero-agent/app/functions/agent-validator/diagnose"
--- a/app/functions/certifik8s/main.go
+++ b/app/functions/certifik8s/main.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	_ "github.com/KimMachineGun/automemlimit"
 	"github.com/cloudzero/cloudzero-agent/app/domain/certificate"
 	"github.com/cloudzero/cloudzero-agent/app/domain/k8s"
 )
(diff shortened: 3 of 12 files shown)


josephbarnett (approved):
Clever :-)

PR #863, opened 2026-06-22
CP-43513: Stamp record timestamps in the streaming store

amfelso:
The backfill job emitted resource metadata metrics (cloudzero_*_labels and cloudzero_*_annotations) with a sample timestamp far in the past instead of the discovery time, corrupting the time attribution of backfilled metadata.

Root Cause:

The backfill path uses the streaming store, whose Create/Update never set RecordCreated/RecordUpdated (unlike the SQLite repo used by the webhook server). The pusher derives each sample's timestamp from maxTime(RecordUpdated, RecordCreated); with both left at the zero value, time.Time{}.UnixNano()/int64(time.Millisecond) serializes as a large negative sentinel that renders as the year 1754. The webhook-server path was unaffected.

Functional Requirements:

1. Backfilled records must carry a real timestamp, matching the webhook path.

   streaming.Store now holds an injected types.TimeProvider and stamps RecordCreated/RecordUpdated on Create (Update delegates to Create), mirroring the SQLite repo. The backfill wiring in functions/webhook/main.go passes the clock already in scope into streaming.New.

Validation:

- Added TestStoreCreateStampsTimestamps: Create and Update stamp from an injected clock.
- Extended the end-to-end streaming test to assert every emitted sample carries a recent timestamp, not the sentinel.
- go test on the streaming package passes; format, lint, and build are clean.

### DIFF
--- a/app/storage/streaming/store_test.go
+++ b/app/storage/streaming/store_test.go
@@ -29,7 +29,6 @@ import (
 	"github.com/cloudzero/cloudzero-agent/app/storage/streaming"
 	"github.com/cloudzero/cloudzero-agent/app/types"
 	"github.com/cloudzero/cloudzero-agent/app/types/mocks"
-	"github.com/cloudzero/cloudzero-agent/app/utils"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


app/storage/streaming/store_test.go
```diff
@@ -92,6 +94,16 @@ func TestStreamingStoreEndToEnd(t *testing.T) {
 	received := sink.timeseries()
 	require.NotEmpty(t, received, "collector should have received timeseries")
 
+	// Samples must carry a recent timestamp, not the zero-time sentinel.
+	const sentinel = int64(-6795364578871)
+	before := clock.GetCurrentTime().Add(-time.Hour).UnixNano() / int64(time.Millisecond)
+	for _, series := range received {
+		for _, sample := range series.Samples {
+			assert.NotEqual(t, sentinel, sample.Timestamp, "sample must not carry the zero-time sentinel")
+			assert.Greater(t, sample.Timestamp, before, "sample timestamp should be recent")
```
evan-cz: This needs some work; instead of getting the current time and testing that the value is *not* a specific value, you should be using the [mock time provider](https://github.com/Cloudzero/cloudzero-agent/blob/develop/app/types/mocks/mock_time_provider.go) and checking that the value *is* exactly what you expect.

evan-cz (approved):
One small nit regarding the test construction, but other than that lgtm

PR #872, opened 2026-06-23
CP-43357: ship prometheus-config-reloader in the agent image

amfelso:
## What & why

Ships the `prometheus-config-reloader` executable **inside** the CloudZero Agent image so the chart no longer pulls it as a separate external image. This removes one of the few external images a default install still pulls — each of which customers in controlled-supply-chain / air-gapped environments must mirror, scan, and maintain.

Implements **CP-43357** (sub-task of CP-28669).

## Approach

- **Built from source**, not copied from the prebuilt upstream image (which ships with known High/Critical CVEs). A new `reloader` stage in `docker/Dockerfile` clones the pinned tag, bumps two vulnerable transitive deps (`x/crypto`, `x/net`), cross-compiles for amd64/arm64, and `COPY`s only the binary into the final image. No upstream source is vendored into the repo.
- **Chart rewired** to the in-image binary for the reload sidecar (deploy + daemonset) and the federated-mode `config-subst` init container. The external image override is **removed entirely** (no tombstones): the agent image and the `/app/prometheus-config-reloader` command are always used. This is a breaking change — `components.prometheusReloader.image` and `configmapReload.prometheus.image` are gone, and because the values schema is strict (`additionalProperties: false`), a values file still setting either key will fail validation on upgrade and must drop those keys first.
- **No shell needed** for `config-subst`: the reloader runs one-shot (`--watch-interval=0`) using its built-in `$(VAR)` substitution. The config placeholder changed `${NODE_NAME}` → `$(NODE_NAME)`; Prometheus relabel back-references like `${1}` are left untouched. Config comments avoid the literal `$(...)` form so the reloader's envsubst doesn't rewrite them or emit spurious "unset variable" warnings.

## Validation

- `grype --fail-on high` on the built binary's SBOM: **0 High/Critical**.
- Built/extracted the binary for **linux/amd64 and linux/arm64** (static, stripped); confirmed the dep bumps are embedded.
- Ran the binary one-shot: `$(NODE_NAME)` substituted, `${1}` preserved, **exit 0 even when the reload target is down** (init-container safe).
- **Live federated install on a dev EKS cluster** (built image, DaemonSet mode): `config-subst` exits 0 with no envsubst warnings; the live Prometheus config shows `$(NODE_NAME)` resolved to each pod's node name with `${1}` relabel refs preserved; reload sidecar reaches `last_reload_successful=1` with 0 apply failures.
- New helm unit test: default reload sidecar (deploy) and `config-subst` (daemonset) both use the agent image + in-image command.
- `helm-lint` passes; golden template renders regenerated.

## Acceptance criteria

- [x] Reloader executable shipped in the agent image; chart no longer pulls it separately
- [x] No upstream source vendored (no submodule, no checked-in source)
- [x] Chart uses in-image binary; external reloader override removed
- [x] CI image scan passes — 0 High/Critical introduced by the reloader binary
- [x] Builds for linux/amd64 and linux/arm64
- [x] Pin + bump procedure documented (`docker/Dockerfile`, `DEVELOPMENT.md`)
- [ ] **kuttl/helm test asserting a real reload — follow-up PR** (intentionally split out)

## Notes for reviewers

- External reloader override is removed rather than retained as a tombstone (`configmapReload.prometheus.image` and `components.prometheusReloader.image` both gone); see the breaking-change note under Approach.
- The reload e2e test (kuttl/helm) lands in a follow-up PR.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -82,6 +82,7 @@ FROM ghcr.io/cloudzero/alloy:v1.17.0 AS alloy
 # (not copy the prebuilt image) to control the toolchain and bump vulnerable
 # deps; only the binary is copied out, no source is vendored.
 FROM base-tools AS reloader
+ARG TARGETPLATFORM
 ARG TARGETOS TARGETARCH
 ARG RELOADER_VERSION
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,12 +1,19 @@
 ARG DEPLOY_IMAGE=scratch
 
+# prometheus-config-reloader version, built from upstream source (see the
+# "reloader" stage below). This is a git tag, which Dependabot does NOT track
+# (it only watches Chart.yaml and Dockerfile FROM lines), so it must be bumped
+# by hand. See DEVELOPMENT.md ("prometheus-config-reloader") for the procedure.
+ARG RELOADER_VERSION=v0.91.0
+
 # Multi-stage Docker build with platform-specific cache optimization:
 # 1. base-tools: Install system packages and tools (cached per platform)
 # 2. dependencies: Download Go modules (cached per platform)
 # 3. builder: Build Go binaries (cached per platform)
 # 4. certs: Extract certificates from distroless image
 # 5. alloy: Extract Alloy binary from CloudZero Alloy image
-# 6. final: Minimal runtime image with compiled binaries
+# 6. reloader: Build prometheus-config-reloader from upstream source
+# 7. final: Minimal runtime image with compiled binaries
 
 # Stage 1: Base tools installation
 FROM --platform=$BUILDPLATFORM golang:1.26.4-alpine AS base-tools
@@ -73,7 +80,36 @@ FROM gcr.io/distroless/static-debian12:debug@sha256:46fcf1fa44d251b0944ba4b98ef4
 # Stage 5: Extract Alloy binary from CloudZero Alloy image
 FROM ghcr.io/cloudzero/alloy:v1.17.0 AS alloy
 
-# Stage 6: Final runtime image
+# Stage 6: Build prometheus-config-reloader from upstream source
+#
+# We build the binary ourselves rather than copying the prebuilt upstream image
+# so that (a) we control the Go toolchain, and (b) we can bump vulnerable
+# transitive dependencies. Building the v0.91.0 tag with our toolchain clears
+# all but a handful of the High/Critical CVEs grype finds in the prebuilt image;
+# bumping x/crypto and x/net below clears the rest so the image scan passes.
+#
+# Only the resulting binary is copied into the final image; no upstream source
+# is vendored into this repository.
+FROM base-tools AS reloader
+ARG TARGETOS TARGETARCH
+ARG RELOADER_VERSION
+
+WORKDIR /reloader-src
+RUN git clone --depth 1 --branch "${RELOADER_VERSION}" \
+        https://github.com/prometheus-operator/prometheus-operator.git .
+
+# Carry these two deps slightly ahead of upstream's pins to clear the remaining
+# grype findings. Remove once upstream pins versions at least this new.
+RUN --mount=type=cache,target=/go/pkg/mod,id=gomod-reloader-$TARGETPLATFORM \
+    go get golang.org/x/crypto@v0.52.0 golang.org/x/net@v0.55.0
+
+RUN --mount=type=cache,target=/go/pkg/mod,id=gomod-reloader-$TARGETPLATFORM \
+    --mount=type=cache,target=/root/.cache/go-build,id=gobuild-reloader-$TARGETPLATFORM \
+    CGO_ENABLED=0 GOOS="$TARGETOS" GOARCH="$TARGETARCH" \
+    go build -trimpath -ldflags="-s -w" \
+        -o /bin/prometheus-config-reloader ./cmd/prometheus-config-reloader
+
+# Stage 7: Final runtime image
 # Note: For debugging, you can temporarily change the image used for building by
 # passing in something like this to 'docker build':
 #
@@ -119,6 +155,9 @@ COPY --from=builder /go/bin/cloudzero-certifik8s /app/cloudzero-certifik8s
 
 # Copy the Alloy binary from the alloy stage
 COPY --from=alloy /bin/cloudzero-alloy /app/cloudzero-alloy
+
+# Copy the prometheus-config-reloader binary built from upstream source
+COPY --from=reloader /bin/prometheus-config-reloader /app/prometheus-config-reloader
 
 # Allow the default ENTRYPOINT from busybox to be the default,
 # however run the app as the default command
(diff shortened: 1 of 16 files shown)


helm/templates/_cm_helpers.tpl
```diff
@@ -509,11 +509,11 @@ Usage: {{ include "cloudzero-agent.prometheus.scrapeCAdvisor" (dict "root" . "sc
...
     # Filter to only the local node
-    # NOTE: ${NODE_NAME} is a placeholder replaced by an init container
-    # with the actual node name using sed. This is necessary because
-    # Prometheus doesn't support environment variable substitution.
+    # NOTE: $(NODE_NAME) is substituted with the node name by the reloader init
+    # container (Thanos-style $(VAR) syntax; relabel refs like ${1} use braces
+    # and are left untouched). Prometheus has no env var substitution of its own.
     - source_labels: [__meta_kubernetes_node_name]
-      regex: ${NODE_NAME}
+      regex: $(NODE_NAME)
```
evan-cz, line 516: Have you tested this thoroughly? This one has me a bit nervous that Claude was hallucinating something...

helm/templates/agent-daemonset.yaml
```diff
@@ -55,16 +55,21 @@ spec:
...
-        # in the prometheus.yml.in file with the actual node name, then use that
-        # processed file in the container.
+        # Prometheus doesn't support env vars in prometheus.yml, so the reloader
+        # binary substitutes $(NODE_NAME) into the processed config consumed by
+        # the server. --watch-interval=0 runs the substitution once and exits.
         - name: config-subst
-          {{- include "cloudzero-agent.generateImage" (dict "defaults" .Values.defaults.image "image" .Values.components.prometheusReloader.image "compat" .Values.configmapReload.prometheus.image) | nindent 10 }}
+          {{- include "cloudzero-agent.generateImage" (dict "defaults" .Values.defaults.image "image" .Values.components.agent.image "compat" .Values.configmapReload.prometheus.image) | nindent 10 }}
+          {{- /* In-image binary runs by path; an external override image runs via
+                 its own entrypoint, so omit command when the override is set. */}}
```
evan-cz: This comment should go away... we don't need a comment explaining what used to be there; comments should document the why of the code as it is, not provide a comparison of what used to exist.

helm/templates/agent-daemonset.yaml
```diff
@@ -55,16 +55,21 @@ spec:
...
+          {{- if not .Values.configmapReload.prometheus.image.repository }}
```
evan-cz: Earlier comments were talking about how those properties are now vestigial, but you use it here? We're moving to an internal prometheus-config-reloader, external is no longer an option.

helm/templates/agent-daemonset.yaml
```diff
@@ -87,7 +92,11 @@ spec:
       containers:
         {{- if .Values.configmapReload.prometheus.enabled }}
         - name: {{ template "cloudzero-agent.name" . }}-{{ .Values.server.name }}-configmap-reload
-          {{- include "cloudzero-agent.generateImage" (dict "defaults" .Values.defaults.image "image" .Values.components.prometheusReloader.image "compat" .Values.configmapReload.prometheus.image) | nindent 10 }}
+          {{- include "cloudzero-agent.generateImage" (dict "defaults" .Values.defaults.image "image" .Values.components.agent.image "compat" .Values.configmapReload.prometheus.image) | nindent 10 }}
+          {{- if not .Values.configmapReload.prometheus.image.repository }}
```
evan-cz: Same thing here. If we're dropping support, let's drop it. This is just unnecessary complexity.

helm/templates/agent-deploy.yaml
```diff
@@ -118,7 +118,13 @@ spec:
...
         # ConfigMap reloader sidecar for Prometheus/Alloy
         - name: {{ template "cloudzero-agent.name" . }}-{{ .Values.server.name }}-configmap-reload
-          {{- include "cloudzero-agent.generateImage" (dict "defaults" .Values.defaults.image "image" .Values.components.prometheusReloader.image "compat" .Values.configmapReload.prometheus.image) | nindent 10 }}
+          {{- include "cloudzero-agent.generateImage" (dict "defaults" .Values.defaults.image "image" .Values.components.agent.image "compat" .Values.configmapReload.prometheus.image) | nindent 10 }}
+          {{- /* In-image binary runs by path; an external override image runs via
+                 its own entrypoint, so omit command when the override is set. */}}
+          {{- if not .Values.configmapReload.prometheus.image.repository }}
+          command:
+            - /app/prometheus-config-reloader
+          {{- end }}
```
evan-cz: Ditto, for both the comment and conditional

evan-cz (changes requested):

helm/templates/_cm_helpers.tpl
amfelso: Good call to flag it — I ran it live on a dev EKS cluster in federated mode, and `config-subst` substitutes `$(NODE_NAME)` to the real node name in the loaded Prometheus config with `${1}` relabel refs left untouched. I also caught that my comments using a literal `$(VAR)` tripped a spurious "unset variable" envsubst warning, so I reworded those. The reload sidecar reloads cleanly too (`last_reload_successful=1`, 0 errors).

docker/Dockerfile
```diff
@@ -73,7 +78,28 @@ FROM gcr.io/distroless/static-debian12:debug@sha256:46fcf1fa44d251b0944ba4b98ef4
 # Stage 5: Extract Alloy binary from CloudZero Alloy image
 FROM ghcr.io/cloudzero/alloy:v1.17.0 AS alloy
 
-# Stage 6: Final runtime image
+# Stage 6: Build prometheus-config-reloader from upstream source. We build it
+# (not copy the prebuilt image) to control the toolchain and bump vulnerable
+# deps; only the binary is copied out, no source is vendored.
+FROM base-tools AS reloader
```
evan-cz, line 84: The reloader stage references `$TARGETPLATFORM` in its BuildKit cache-mount IDs (`gomod-reloader-$TARGETPLATFORM`, `gobuild-reloader-$TARGETPLATFORM`) but never declares `ARG TARGETPLATFORM` — and predefined platform args don't carry across `FROM`, so it expanded to empty and both amd64/arm64 shared one cache id (the other three build stages declare it and key per-arch). It's not a correctness bug — `GOOS`/`GOARCH` still drive the actual cross-compile — just cache contention/inconsistency, but can be fixed by adding `ARG TARGETPLATFORM` line right after this line.

evan-cz (approved):
One very minor thing (which doesn't impact correctness, just improves cache performance a bit). Otherwise LGTM

docker/Dockerfile
amfelso: Good catch — fixed in 241feae. Added `ARG TARGETPLATFORM` right after the `FROM` in the reloader stage, matching the base-tools/dependencies/builder stages, so the cache-mount ids key per-arch again instead of sharing one empty-suffixed id.

PR #898, opened 2026-07-03
fix(affinity): update affinity on server and aggregator to be automatically set

pinkneyj:
# Why?

> Affinity is not automatically set for aggregator and server. 

# What

> This change does this, based on code already in webhook-deploy.yaml Although changes have recently been made to set aggregator affinity (https://github.com/Cloudzero/cloudzero-agent/pull/881) and the server affinity can already be set, this just removes the need to do that yourself.
This is also a copy of: https://github.com/Cloudzero/cloudzero-charts/pull/224 but believe this is the repo the change should be made in.

# How Tested

> Deployed locally. Gives values as expected and can be overridden.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -6,7 +6,7 @@
 
 A Helm chart for deploying a metrics collector to send cluster metrics to the CloudZero platform.
 
-For the latest release, see [Releases](https://github.com/Cloudzero/cloudzero-charts/releases). You can also [enable release notifications](#enabling-release-notifications).
+For the latest release, see [Releases](https://github.com/Cloudzero/cloudzero-charts/releases). You can also [enable release notifications](#enabling-release-notifications). 
 
 ## Prerequisites
 
--- a/helm/README.md
+++ b/helm/README.md
@@ -6,7 +6,7 @@
 
 A Helm chart for deploying a metrics collector to send cluster metrics to the CloudZero platform.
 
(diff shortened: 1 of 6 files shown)


pinkneyj:
> Hi @pinkneyj , thank you again for another contribution! This is definitely a good change and I would certainly like to get this behavior merged in, one way or another. The anti-affinity term looks good to me, and you've followed [[the pattern that already exists in the webhook-server template](https://github.com/Cloudzero/cloudzero-agent/blob/develop/helm/templates/webhook-deploy.yaml#L106-L120)](https://github.com/Cloudzero/cloudzero-agent/blob/develop/helm/templates/webhook-deploy.yaml#L106-L120).
> 
> But, your change has surfaced an opportunity for improvement in the chart generally. In the existing templates and in the proposed change, we're baking in an opinionated setting for pod affinity that cannot be removed or overridden. It's a reasonable default behavior, but we also want to allow users the flexibility to choose. What I would propose instead is to move this default - for the agent-deploy, the aggregator-deploy, and the webhook-server deploy - into the values.yaml file, declared on each component's own affinity key. This would mean that:
> 
> 1. All components get the default affinity setting.
> 2. Users have better visibility into the setting.
> 3. Users can remove, override, or alter the setting.
> 
> Let me know what you think. If you're up for the change in this PR, then that's great. Otherwise, we can handle it on our side. Thank you again for the PR!

Hey @dmepham thanks for this response. I've made some changes based on your suggestions, do have a look when is convenient to do so.

Thanks in advance!

PR #944, opened 2026-08-10
CP-45941: Upgrade RELOADER_VERSION to v0.93.0

dmepham:
# Why?

Resolves security build check failures

# What

# How Tested


evan-cz (approved):
This is okay, but the title should be about upgrading prometheus-config-reloader, not an implementation detail like the `RELOADER_VERSION` variable...

